PR #3, opened 2016-03-13
Trivial audit fixes

rwhogg:
This is just a collection of very trivial audit fixes. It doesn't make them all compliant, just more so.


maxim-belkin:
Thank you!

Issue #2, opened 2016-03-12
Configure flags, dependencies


maxim-belkin:
TODO:
Check if we need all the flags used by configure scripts and dependencies on `autoconf`.
In particular check if formulas need:
- `--disable-debug`
- `--disable-dependency-tracking`
- `--disable-silent-rules`
- `autoconf` dependency
- `pkg-config` dependency

<hr>

_[undone on March 14]_ Done ==> For all dynamic libraries run `ldd <library>` and check that emerging dependencies appear in the formula as "runtime" libraries.

``` bash
### in ./Library/Taps/linuxbrew/homebrew-xorg folder
for each in *.rb; do
    name=${each%.rb};
    unset a;
    echo -n "Formula: $name, ";
    a=($(brew list $name -v | grep "\.so$"));
    n=${#a[@]};
    echo "$n dependencies";
    for ai in ${a[@]}; do
       ldd $ai | grep -v -e "linux-vdso\.so\.1" -e "ld-linux-x86-64\.so\.2" -e "libc\.so\.6" -e "libdl\.so\.2" | awk '{print "    ",$1}';
    done | sort -u
done | tee deps.txt
```

<hr>

How to make the following kind of dependency?

``` ruby
# pseudo-code:
"depends_on "libxdmcp"  => :run if ! Formula["libxcb"].buildwithout?("libxdmcp")
```

bw-yuhogg:
I'd be happy to help tomorrow with checking the formulae for excess `configure` flags.

maxim-belkin:
That'd be great! Thanks! :)

sjackman:
> `depends_on "libxdmcp"  => :run if ! Formula["libxcb"].buildwithout?("libxdmcp")`

There's no need to tag a dependency explicitly as `:run`. It's only used, as far as I know, to silence warnings about dependencies that are usually build time dependencies, like `cmake`, but for some particular formula is exceptionally also a run-time dependency.

Perhaps something like this? like here: https://github.com/Linuxbrew/linuxbrew/blob/master/Library/Formula/python.rb#L34

``` ruby
depends_on "libxdmcp"  => if Tab.for_name("libxcb").with?("libxdmcp")
```

I don't think you want to list these recursive dependencies typically though. Just list

``` ruby
depends_on "libxcb"
```

and then in `libxcb.rb`

``` ruby
depends_on "libxdmcp" => :recommended
```

[...]

PR #5, opened 2016-03-13
[WIP] additional configure flags

rwhogg:
See #2 .

All of the formulae tried so far accept `--disable-silent-rules` and `--disable-dependency-tracking`, but not `--disable-debug`.
## TODO
- [ ] Rebuild all the formulae from scratch once I've updated them all.

### DIFF
--- a/xcb-proto.rb
+++ b/xcb-proto.rb
@@ -25,6 +25,7 @@ class XcbProto < Formula
       --prefix=#{prefix}
       --sysconfdir=#{etc}
       --localstatedir=#{var}
+      --disable-silent-rules
     ]
 
     system "./configure", *args
--- a/bigreqsproto.rb
+++ b/bigreqsproto.rb
@@ -19,6 +19,8 @@ class Bigreqsproto < Formula
       --prefix=#{prefix}
       --sysconfdir=#{etc}
       --localstatedir=#{var}
+      --disable-silent-rules
+      --disable-dependency-tracking
     ]
 
     system "./configure", *args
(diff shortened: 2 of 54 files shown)


maxim-belkin:
Cool! Thanks! Does `xorg` build fine with these rules?

bw-yuhogg:
Haven't been able to check that yet. I'll make sure to do so before removing the "WIP" from the title. (Adding a TODO to the summary to remind myself)

It is _really_ tedious to do this because of how many libraries there are, so it may take a while.

xcb-proto.rb
```diff
@@ -25,6 +25,7 @@ def install
       --prefix=#{prefix}
       --sysconfdir=#{etc}
       --localstatedir=#{var}
+      --disable-silent-rules
```
bw-yuhogg, line 28: Yes, this formula takes `--disable-silent-rules` but not `--disable-dependency-tracking`.

maxim-belkin:
Alright. I'm going to merge these and run a test suite on the updated formulas (I've automated it via a script).

bw-yuhogg:
This was not quite done yet, there were still some more formulae I had yet to check. 

On March 14, 2016 12:11:31 PM PDT, Maxim Belkin notifications@github.com wrote:

> Merged #5.
> 
> ---
> 
> Reply to this email directly or view it on GitHub:
> https://github.com/Linuxbrew/homebrew-xorg/pull/5#event-589229579

## 

Sent from my Android device with K-9 Mail. Please excuse my brevity.

maxim-belkin:
Now you can push, so commit whenever you are done with any single formula.

Side note: I've removed a dependency on autoconf from the formulas as we don't actually need it. I believe it is needed only when `autoconf` or `autoreconf` are used. I'm testing the change right now.

PR #49, opened 2016-05-07
xorg-protocols: Build a bottle for Linuxbrew

rwhogg:


bw-yuhogg:
From https://travis-ci.org/Linuxbrew/homebrew-xorg/builds/128445441#L954-L956

```
Error: 1 problem in 1 formula
linuxbrew/xorg/xorg-protocols:
  * Stable: version (latest) is set to a string without a digit
```

Just an audit failure. Not concerned.

PR #57, opened 2016-05-07
libxmu: Build a bottle for Linuxbrew

rwhogg:


bw-yuhogg:
Might need to redo this bottle since it depends on `libxt` as found at #70 

Other than `libxt` itself though, this should be the only problematic bottle, so minimal damage.

PR #60, opened 2016-05-07
libxfixes: Build a bottle for Linuxbrew

rwhogg:


bw-yuhogg:
Just an audit failure, not concerned.

PR #61, opened 2016-05-07
libfontenc: Depends on zlib

rwhogg:


bw-yuhogg:
From https://travis-ci.org/Linuxbrew/homebrew-xorg/builds/128505817#L563-L567

```
checking zlib.h usability... no
checking zlib.h presence... no
checking for zlib.h... no
configure: error: in `/tmp/libfontenc-20160507-3878-14u3fk2/libfontenc-1.1.3':
configure: error: zlib.h is required to compile libfontenc
```

I think there's a missing dependency on `zlib` :laughing:

PR #70, opened 2016-05-08
libxaw: Build a bottle for Linuxbrew

rwhogg:


bw-yuhogg:
From https://travis-ci.org/Linuxbrew/homebrew-xorg/builds/128643294#L2300-L2303

```
checking for XAW6... configure: error: Package requirements (xproto x11 xext xextproto xt xmu) were not met:
Package 'sm', required by 'xt', not found
```

Need to carefully check the dependencies of this library. Not merging for now.

maxim-belkin:
interesting. will check.

maxim-belkin:
I've noticed a few weird things with the gist:
1. Installation doesn' see `util-macros`, though it is requested. Also, it is not in the list of uninstalled packages in the very end. Could you please check what's going on there?
2. `libsm` is in the list of dependencies, so... it should be installed. I wonder if Travis changes the order of dependencies somehow? But in any way, `libsm` should come begore `libxaw`.

maxim-belkin:
Just tried installing `libxaw` from scratch on my machine and installation succeeded.

bw-yuhogg:
``` bash
$ patchelf --print-needed ~/.linuxbrew/opt/libxaw/lib/libXaw7.so
libXext.so.6
libXmu.so.6
libXt.so.6
libXpm.so.4
libX11.so.6
libc.so.6
$ brew deps --1 libxaw
libx11
libxext
libxmu
libxpm
libxt
```

`libsm` isn't a direct dependency, so it doesn't need to be added to this package. From the looks of things, `libxaw` has the correct dependencies declared.

The problem is, I think, in `libxt`.

``` bash
$ patchelf --print-needed ~/.linuxbrew/opt/libxt/lib/libXt.so
libSM.so.6
libICE.so.6
libX11.so.6
libc.so.6
$ brew deps --1 libxt 
libice
libx11
```

The problem appears to be that `libsm` is a _build_ dependency of `libxt`. @maxim-belkin, was that just a typo?

> 1. Installation doesn' see util-macros, though it is requested. Also, it is not in the list of uninstalled packages in the very end. Could you please check what's going on there?

Note to self: remember to do that.

maxim-belkin:
> The problem appears to be that `libsm` is a build dependency of `libxt`. @maxim-belkin, was that just a typo?

That's a mistake. Thanks! I've fixed it in 34c6e5b

PR #79, opened 2016-05-09
libxtst: Build a bottle for Linuxbrew

rwhogg:


bw-yuhogg:
```
checking for XTST... configure: error: Package requirements (x11 xext >= 1.0.99.4 xi recordproto >= 1.13.99.1 xextproto >= 7.0.99.3 inputproto) were not met:
Package 'xfixes', required by 'xi', not found
```

`libxi` has `libxfixes` as a `:build` dependency, but it doesn't appear to be actually linked against

``` bash
$ patchelf --print-needed ~/.linuxbrew/opt/libxi/lib/libXi.so
libX11.so.6
libXext.so.6
libc.so.6
```

Or even indirectly linked against

``` bash
$ ldd ~/.linuxbrew/opt/libxi/lib/libXi.so
    linux-vdso.so.1 (0x00007ffe70758000)
    libX11.so.6 => /home/bob/.linuxbrew/lib/libX11.so.6 (0x00007fdbcf601000)
    libXext.so.6 => /home/bob/.linuxbrew/lib/libXext.so.6 (0x00007fdbcf3f0000)
    libc.so.6 => /home/bob/.linuxbrew/lib/libc.so.6 (0x00007fdbcf078000)
    libxcb.so.1 => /home/bob/.linuxbrew/lib/libxcb.so.1 (0x00007fdbcee57000)
    libdl.so.2 => /home/bob/.linuxbrew/lib/libdl.so.2 (0x00007fdbcec53000)
    /home/bob/.linuxbrew/Cellar/glibc/2.19/lib64/ld-linux-x86-64.so.2 (0x00007fdbcfb2c000)
    libXau.so.6 => /home/bob/.linuxbrew/lib/libXau.so.6 (0x00007fdbcea4f000)
    libXdmcp.so.6 => /home/bob/.linuxbrew/lib/libXdmcp.so.6 (0x00007fdbce848000)
```

Hmm...

bw-yuhogg:
From `libxi`'s changelog:

```
    Require XFixes for PointerBarrier, remove duplicate typedef

    The PointerBarrier typedef is duplicate if a client includes both Xfixes.h
    and XInput2.h.
```

So it looks like this is in fact a build-time dependency.

bw-yuhogg:
``` bash
$ brew uses libxfixes
linuxbrew/xorg/libxdamage                linuxbrew/xorg/libxcursor                linuxbrew/xorg/xorg
$ brew uses libxi
linuxbrew/xorg/libxtst                                       linuxbrew/xorg/xorg
$ patchelf --print-needed ~/.linuxbrew/lib/libXtst.so
libX11.so.6
libXext.so.6
libXi.so.6
libc.so.6
```

I think we have three options here. One is to make `libxi`'s dependency on `libxfixes` be normal rather than `:build`, so that `libxtst` can count on it being there; one is to make `libxtst` depend on `libxfixes` (presumably as a `:build`), and the final one is that this is actually an upstream bug.

From `libxtst`'s Changelog:

```
Add xextproto & xi to Requires.private

    Only headers are needed, not libraries to link with.   Required for
    includes of xtestconst.h (xextproto) and XInput.h (xi) in XTest.h
    public header.

    Fixes https://bugs.freedesktop.org/attachment.cgi?id=59835
    (originally reported as https://bugzilla.novell.com/show_bug.cgi?id=748808 )
```

The case for it being an upstream bug would be that since `libxfixes` provides a structure in its header file that `libxi` used to provide itself, `libxi`'s pkg-config file should have `libxfixes` as _Requires_, not _Requires.private_. However, I'm not an expert on pkg-config, so I'm not sure if that's how that should work.

@sjackman, @maxim-belkin, opinions?

maxim-belkin:
I think we should go the second route and add a comment in the formula that this is an upstream bug.

sjackman:
If the goal was to install only the libraries need to run (not compile) X programs, then I would go with "one is to make libxtst depend on libxfixes (presumably as a :build)". The downside is that you'll have to do that for every formula that depends on `libxi`, and it won't work for providing a system to the user that can be used to compile X programs. The simplest solution is to change the `:build` dependency to normal. I have a slight preference for this solution, but not strongly held. Either will work.

maxim-belkin:
Wait a second: I think `libxtst` is missing a dependency on `util-macros`...
@rwhogg: try adding it as a build dependency and building a bottle once again.

bw-yuhogg:
Sadly, that doesn't look sufficient. After installing `libxi` on my machine and removing `libxfixes`, building with the `util-macros` build dependency gives the same error as Travis gets.

bw-yuhogg:
I've opened #83 to make `libxfixes` a normal dependency instead of a build dependency.

Since this has bitten us twice now, I think we ought to take a look at the remaining build dependencies and consider changing them to normal dependencies. After all, most users are going to have _almost all_ the formula from this tap installed, assuming they've tapped it at all, by virtue of installing `linuxbrew/xorg/xorg`.

``` bash
$ ag "depends_on.*lib.*build" | grep -v libxslt | grep -v glib | grep -v zlib | grep -v libxml2 | grep -v libtool | grep -v libxi.rb
libxft.rb:21:  depends_on "libxrender" =>  :build
libxrandr.rb:18:  depends_on "libxrender" =>  :build
libdrm.rb:15:  depends_on "libpciaccess" => :build
libxcomposite.rb:19:  depends_on "libxfixes"  =>  :build
```

sjackman:
`egrep -v 'libxslt|glib|zlib|libxml2|libtool|libxi'` Less typing. =)

PR #90, opened 2016-05-13
libxshmfence: be explicit about args

rwhogg:


bw-yuhogg:
make check failed: `FAIL: xshmfence_test`

sjackman:
It passes on LinuxbrewTestBot on `ubuntu:trusty`. Just an audit failure.

```
==> brew audit linuxbrew/xorg/libxshmfence
Error: 1 problem in 1 formula
linuxbrew/xorg/libxshmfence:
  * Don't negate 'build.without?': use 'build.with?'
==> FAILED
```

bw-yuhogg:
Yeah, it passes on my system too. I haven't had time to debug this yet (been really busy recently) but did not like the idea of merging with a test failure on Travis.

bw-yuhogg:
Had to debug it locally in the `linuxbrew/linuxbrew` Docker image. Found this:

```
FAIL: xshmfence_test
====================

xshmfence_alloc_shm: Permission denied
```

bw-yuhogg:
From https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=773724

```
> FAIL: xshmfence_test
> ====================
> 
> xshmfence_alloc_shm: Permission denied
> 
Fix your setup, /dev/shm needs to be world-writable.
```

That didn't work, but making `/run/shm` world-writable did. (Thanks a bunch to Julien Cristau for the suggestion though!)

bw-yuhogg:
Note to @maxim-belkin: I set the X shared memory directory to be $HOMEBREW_PREFIX/var/run/shm so that we can guarantee it's world-writable (decided this was preferable to forcing every system/Docker container to have to change its permissions).

Do you anticipate any problems with this?

maxim-belkin:
Looks good to me. Please also change "with-static" part to the following:
`args << "--enable-static=#{build.with?("static") ? "yes" : "no"}"`

sjackman:
Seems like a Travis bug to me that `/dev/shm` isn't world-writeable. It should be correct on any normal distro. I'd prefer to keep the default directory of `/dev/shm` or `/run/shm`, and perhaps add a `brew doctor` check if a user reports this issue.

bw-yuhogg:
[This Docker ticket](https://github.com/docker/docker/issues/5126) is for `/dev/shm` being mounted 755 instead of 777, but it looks like it was fixed two years ago. As for `/run/shm`, couldn't find one for that.

I think we can probably fix `/run/shm` with a change to the Dockerfile.

At any rate, undoing the change to the directory.

Issue #11, opened 2016-03-20
xorg-docs: I/O error : Attempt to load network entity


sjackman:
https://gist.github.com/sjackman/9ddb225ae6eda933a6cb#file-02-make-L14

```
/gsc/btl/linuxbrew/bin/xsltproc --path "/gsc/btl/linuxbrew/Cellar/xorg-sgml-doctools/1.11/share/sgml/X11" --path "/var/tmp/sjackman/xorg-docs20160319-115225-tk516u/xorg-docs-1.7" --stringparam targets.filename "fonts.pdf.db" --stringparam collect.xref.targets "only" --stringparam olink.base.uri "fonts.pdf" --nonet --xinclude /gsc/btl/linuxbrew/Cellar/xorg-sgml-doctools/1.11/share/sgml/X11/xorg-fo.xsl fonts.xml
I/O error : Attempt to load network entity http://docbook.sourceforge.net/release/xsl/current/fo/docbook.xsl
warning: failed to load external entity "http://docbook.sourceforge.net/release/xsl/current/fo/docbook.xsl"
compilation error: file /gsc/btl/linuxbrew/Cellar/xorg-sgml-doctools/1.11/share/sgml/X11/xorg-fo.xsl line 11 element import
xsl:import : unable to load http://docbook.sourceforge.net/release/xsl/current/fo/docbook.xsl
Makefile:529: recipe for target 'fonts.pdf.db' failed
make[2]: *** [fonts.pdf.db] Error 5
```

maxim-belkin:
Hmm, are you saying that `xorg-docs` tries to load something from `sourceforge.net` and that sourceforge may be down and, therefore, the installation fails when it happens? We could host this file as a part of the package and use `inreplace` to redirect to our location.

I don't have access to my Linux workstation right now (network issues...).

sjackman:
I haven't had a chance to troubleshoot the error. I'm not sure if it's a transient error. It's an unusual error message. I'll try it again tomorrow and see whether it succeeds. You can just ignore it for now. Does `brew install xorg-docs` work for you?

sjackman:
Still failing for me today.

maxim-belkin:
Just to make sure: what happens when you enter `http://docbook.sourceforge.net/release/xsl/current/fo/docbook.xsl` into the url bar of a browser?

sjackman:
That works.

DoomHammer:
Had the same problem with `docbook`.

[...]

PR #94, opened 2016-05-14
xorg-docs: build a bottle for Linuxbrew

rwhogg:


bw-yuhogg:
While building `libxslt`:

```
/usr/bin/ld: cannot find -lz
collect2: error: ld returned 1 exit status
```

Ah man. I just merged the `libxslt` bottle :frowning:

bw-yuhogg:
I've traced the `zlib` link to `libxml2`. This is kind of weird though. From [`libxml2`'s website](http://xmlsoft.org/FAQ.html):

```
What other libraries are needed to compile/install libxml2 ?

Libxml2 does not require any other library, the normal C ANSI API should be sufficient (please report any violation to this rule you may find).

However if found at configuration time libxml2 will detect and use the following libs:

    libz : a highly portable and available widely compression library.
    iconv: a powerful character encoding conversion library. It is included by default in recent glibc libraries, so it doesn't need to be installed specifically on Linux. It now seems a part of the official UNIX specification. Here is one implementation of the library which source can be found here.
```

@sjackman, I'm going to add a `depends_on "zlib" => :recommended unless OS.mac?` to `libxml2`. We can worry about `libiconv` later, since it doesn't build correctly at the moment.

bw-yuhogg:
Either way though, this isn't actually the major problem here. The major problem is that it can't find various external entities:

```
I/O error : Attempt to load network entity http://www.oasis-open.org/docbook/xml/4.3/docbookx.dtd
warning: failed to load external entity "http://www.oasis-open.org/docbook/xml/4.3/docbookx.dtd"
validity error : Could not load the external subset "http://www.oasis-open.org/docbook/xml/4.3/docbookx.dtd"
Document /tmp/xorg-docs-20160514-22943-bh592y/xorg-docs-1.7/general/fonts/fonts.xml does not validate
```

It also complains in several places that it doesn't know what `nbsp` means.

bw-yuhogg:
We're running into #11 here. We can worry about this some other time; this isn't blocking the X11 requirement.

bw-yuhogg:
```
Error: JAVA_HOME is not defined correctly.
  We cannot execute java
/tmp/xmlto.eXNzwr
Error: JAVA_HOME is not defined correctly.
```

While this isn't a major problem, it reveals that `fop` depends on Java. I think I'll submit that upstream.

bw-yuhogg:
> While this isn't a major problem, it reveals that fop depends on Java. I think I'll submit that upstream.

Done and merged in https://github.com/Homebrew/homebrew-core/pull/1330

sjackman:
``` ruby
ENV["XML_CATALOG_FILES"] = etc/"xml/catalog"
```

Does this variable need to set in the user's environment as well? If so, should it be documented in caveats of the relevant formula?

sjackman:
`linuxbrew/linuxbrew/xorg-docs` has a `:build` dependency on `fop` which depends on `:java` which is satisfied by `jdk` on Linuxbrew. Installing the `xorg-docs` bottle pulls in `jdk` as a dependency even though the dependency on `fop` is a `:build` dependency. This I believe is a bug in the dependency resolver of `brew`. @xu-cheng 

```
$ brew install linuxbrew/linuxbrew/xorg-docs
==> Installing xorg-docs from linuxbrew/xorg
==> Installing dependencies for linuxbrew/xorg/xorg-docs: jdk
==> Installing linuxbrew/xorg/xorg-docs dependency: jdk
==> Downloading http://download.oracle.com/otn-pub/java/jdk/8u60-b27/jdk-8u60-li
==> Downloading from http://download.oracle.com/otn-pub/java/jdk/8u60-b27/jdk-8u
######################################################################## 100.0%
==> Caveats
By installing and using JDK you agree to the
Oracle Binary Code License Agreement for the Java SE Platform Products and JavaFX
http://www.oracle.com/technetwork/java/javase/terms/license/index.html

==> Summary
🍺  /home/sjackman/.linuxbrew/Cellar/jdk/1.8.0-60: 1,622 files, 348.0M, built in 1 minute 17 seconds
==> Installing linuxbrew/xorg/xorg-docs
==> Downloading https://linuxbrew.bintray.com/bottles-xorg/xorg-docs-1.7.x86_64_
######################################################################## 100.0%
==> Pouring xorg-docs-1.7.x86_64_linux.bottle.tar.gz
🍺  /home/sjackman/.linuxbrew/Cellar/xorg-docs/1.7: 136 files, 7.6M
```

bw-yuhogg:
``` bash
$ brew deps xorg-docs
fop
jdk
libxml2
libxslt
util-macros
zlib
$ brew deps --include-build xorg-docs
bzip2
docbook
docbook-xsl
fop
gettext
gnu-getopt
gpatch
homebrew/dupes/ncurses
homebrew/dupes/unzip
jdk
libxml2
libxslt
pkg-config
util-macros
xmlto
xorg-sgml-doctools
xz
zlib
$ brew info xorg-docs
...
==> Dependencies
Build: util-macros ✔, xmlto ✔, fop ✔, libxslt ✔, xorg-sgml-doctools ✔, docbook ✔, docbook-xsl ✔, gpatch ✔
Recommended: util-macros ✔, fop ✔, libxslt ✔
...
```

Worth noting that the dependency on `fop` is `depends_on "fop" => [:build, :recommended]` whereas the dependency on `docbook` is `depends_on "docbook" => :build` (and likewise for `docbook-xsl` and `xorg-sgml-doctools`). It looks like `:recommended` is overruling `:build`.

That might or might not be a bug. It might just mean that we're misinterpreting what `:recommended` means.

maxim-belkin:
What we might need to do is:

```
option "without-fop", "Build without fop"
depends_on "fop" => :build
```

bw-yuhogg:
> ENV["XML_CATALOG_FILES"] = etc/"xml/catalog"
> Does this variable need to set in the user's environment as well? If so, should it be documented in caveats of the relevant formula?

Found this in `asciidoc`'s caveats (not the same obviously, but a useful guide if nothing else):

```
      If you intend to process AsciiDoc files through an XML stage
      (such as a2x for manpage generation) you need to add something
      like:

        export XML_CATALOG_FILES=#{etc}/xml/catalog

      to your shell rc file so that xmllint can find AsciiDoc's
      catalog files.

      See `man 1 xmllint' for more.
```

`xmllint`'s man page has this to say:

```
       XML_CATALOG_FILES
           XML catalog behavior can be changed by redirecting queries to the user's own set of catalogs. This can be
           done by setting the XML_CATALOG_FILES environment variable to a list of catalogs. An empty one should
           deactivate loading the default /etc/xml/catalog catalog.
```

However, people might have important XML DTD's listed in `/etc/xml/catalog` and appearing somewhere in `/etc/xml`. Catalog entries look like this: `<delegateSystem systemIdStartString="http://docutils.sourceforge.net/" catalog="file:///etc/xml/docutils-common.xml"/>`.

I'm not really sure what would happen if two catalog files pointed the same URL to two conflicting DTD's, but, assuming it works like PATH and the first catalog file with a hit is used, I think for most users the correct thing to do would be something like `export XML_CATALOG_FILES=$HOMEBREW_PREFIX/etc/xml/catalog:/etc/xml/catalog`, assuming that the second location is constant (or nearly so) across all Linux distros. Nevertheless, given that it potentially applies to a number of XML-related formulae, I'm not sure that any formula's caveats in particular is a good place to put it.

Thoughts?

PR #96, opened 2016-05-22
encodings: break circular dependency between libfontenc and mkfontscale

rwhogg:
THIS IS A TEST!

### DIFF
--- a/encodings.rb
+++ b/encodings.rb
@@ -9,8 +9,7 @@ class Encodings < Formula
   # tag "linuxbrew"
 
   depends_on "pkg-config" =>  :build
-  depends_on "font-util"  =>  :build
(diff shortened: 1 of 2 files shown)


bw-yuhogg:
Okay, so `encodings` definitely depends on `mkfontscale`.

`brew install mkfontscale --ignore-dependencies` gives this:

```
checking for MKFONTSCALE... configure: error: Package requirements (fontenc freetype2) were not met:
No package 'fontenc' found
```

`brew install -v libfontenc --ignore-dependencies` works though, so it looks like that may be the one that has an excess dependency.

bw-yuhogg:
@maxim-belkin fixed this in 946c51958e9d9350ad712bdfd290c8626a7ecf62

PR #102, opened 2016-05-25
font-adobe-100dpi: Build a bottle for Linuxbrew

rwhogg:


bw-yuhogg:
This (and a bunch of the other font formulae) are failing because of bdftopcf. I've opened a separate PR for that at #108 .

Marking as blocked for now.

maxim-belkin:
This is not due to `bdftopcf`. The error message says:

```
checking for BDFTOPCF... configure: error: Package requirements (xfont) were not met:
Package 'xproto', required by 'xfont', not found
```

`xproto` is listed as ":build" dependency of `libxfont`. So, making it a standard dependency should work then.

maxim-belkin:
The same is true for `fontsproto` dependency. I've committed both fixes.
Other dependencies need to be checked as well.
The way to do it is like so:
Parse `.pc` file of a package, e.g. `[linuxbrew]/Cellar/libxfixes/5.0.1/lib/pkgconfig/xfixes.pc` and find a line that starts with `Requires` and make found dependencies standard dependencies of the package.

maxim-belkin:
There was also a missing "run-time" dependency on `mkfontscale` in `mkfontdir`. Interestingly, `mkfontdir` is simply a script that calls `mkfontscale` but doesn't require it while being built

PR #103, opened 2016-05-25
font-adobe-75dpi: Build a bottle for Linuxbrew

rwhogg:


bw-yuhogg:
This (and a bunch of the other font formulae) are failing because of bdftopcf. I've opened a separate PR for that at #108 .

Marking as blocked for now.

PR #138, opened 2016-05-31
libsha1: Depends on automake

rwhogg:


bw-yuhogg:
```
Can't exec "aclocal": no such file or directory
```

Missing a build dep on `automake`.

PR #167, opened 2016-07-26
xhost: Build a bottle for Linuxbrew

rwhogg:


bw-yuhogg:
Failed because of audit problems. Fixed separately and now rebuilding.

PR #191, opened 2016-07-31
font-bitstream-type1: Build a bottle for Linuxbrew

maxim-belkin:


maxim-belkin:
Closing this as bottle is not `cellar :any` and, actually, empty.

PR #168, opened 2016-07-27
Mesa: update version to 13.0.4

maxim-belkin:
## Contributing to homebrew-xorg:

_You can erase any parts of this template not applicable to your Pull Request._
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Linuxbrew/homebrew-xorg/blob/master/.github/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Linuxbrew/homebrew-xorg/pulls) for the same update/change?
- [x] Does your submission pass
  `brew audit --strict <formula>` (where `<formula>` is the name of the formula you're submitting)?
- [x] Have you built your formula locally prior to submission with `brew install <formula>`?

PR #193, opened 2016-08-04
libpthread-stubs 0.3

ntkme:
This PR provides libpthread-stubs 0.3.

There are three packages depend on this right now: libdrm, libxcb, mesa (#168). The current solution is [patching the code to skip checking for `libpthread-stubs`](https://github.com/Linuxbrew/homebrew-xorg/search?utf8=✓&q=libpthread-stubs). It took me a while to figure it out. So, I think including this package will reduce the future confusion for other contributors.

I understand that this package effectively does nothing on Linux, but I don't see any side effect having it. Plus, contributors can save some time on patching files.

---
#### $prefix/lib/pkgconfig/pthread-stubs.pc

```
prefix=/home/linuxbrew/.linuxbrew/libpthread-stubs/0.3
exec_prefix=${prefix}
libdir=${exec_prefix}/lib

Name: pthread stubs
Description: Stubs missing from libc for standard pthread functions
Version: 0.3
Libs:
```

### DIFF
--- a/libxcb.rb
+++ b/libxcb.rb
@@ -14,13 +14,14 @@ class Libxcb < Formula
   option "with-static", "Build static libraries (not recommended)"
   option "with-docs",   "Generate API documentation"
 
-  depends_on "pkg-config"  => :build
+  depends_on "pkg-config" => :build
   depends_on "libxau"
-  depends_on "xcb-proto"   => :build
-  depends_on "libxdmcp"    => :recommended
+  depends_on "xcb-proto" => :build
+  depends_on "libpthread-stubs" => :build
+  depends_on "libxdmcp" => :recommended
 
   depends_on "doxygen" => :build if !build.with?("docs")
-  depends_on "check"   => :build if build.with?("test")
+  depends_on "check" => :build if build.with?("test")
   depends_on "libxslt" => [:build, :optional]
 
   def install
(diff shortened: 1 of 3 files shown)


maxim-belkin:
I'm actually fine with this. Please add the following `test` block:

```
  def test
    assert File.exist?("#{lib}/pkgconfig/pthread-stubs.pc") 
  end 
```

From README:

> On platforms which already supply all the necessary pthread stubs, this
>  package ships only the pkg-config file pthread-stubs.pc, to allow
>  libraries to unconditionally express a dependency on pthread-stubs and
>  still obtain correct behavior.

libxcb.rb
```diff
@@ -14,10 +14,11 @@ class Libxcb < Formula
   option "with-static", "Build static libraries (not recommended)"
   option "with-docs",   "Generate API documentation"
 
-  depends_on "pkg-config"  => :build
+  depends_on "pkg-config"       => :build
```
bw-yuhogg: Please don't line up the `=>` operators.

I realize that a lot of the existing code does do do that, but I would prefer it if it didn't. My apologies - I hope it didn't take too long to do!

ntkme: Done! I actually have the same preference with you. 😄

libxcb.rb
```diff
@@ -1,7 +1,7 @@
 class Libxcb < Formula
   desc "An interface to the X Window System protocol, which replaces the current Xlib interface"
-  homepage "http://www.x.org/" ### http://www.linuxfromscratch.org/blfs/view/svn/x/x7lib.html
-  url    "http://xcb.freedesktop.org/dist/libxcb-1.11.1.tar.bz2"
```
bw-yuhogg, line 4: :+1:

maxim-belkin:
Thanks a lot, @ntkme !

PR #195, opened 2016-08-05
libxcb: fix dependency on doxygen

ntkme:
This PR fixes a typo.

In #193, I fixed the audit error `Don't negate 'build.without?': use 'build.with?'` by reversing the syntax without paying attention to the actual logic. I looked at it again and noticed that the extra `!` in the original version was actually a typo, which makes that the package depends on doxygen only when docs is _not_ selected...


maxim-belkin:
I think this is correct! Merging.

PR #220, opened 2016-11-11
Various audit fixes

rwhogg:

This fixes the audits for a number of formulae.

I didn't get all the fixes necessary, but I got as many as I felt reasonable to do at once.

### DIFF
--- a/freeglut.rb
+++ b/freeglut.rb
@@ -1,5 +1,5 @@
 class Freeglut < Formula
-  desc "An open-source clone of the GLUT library"
+  desc "Open-source clone of the GLUT library"
   homepage "http://freeglut.sourceforge.net/"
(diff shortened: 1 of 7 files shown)


bw-yuhogg:
Various errors from Travis:
- `configure: error: C compiler cannot create executables`

The problem here appears to be with GCC. To speed up this build, I think we should probably bottle `check` in core, but that wouldn't solve the problem for `libice`.

bw-yuhogg:
`check` has been bottled in https://github.com/Linuxbrew/homebrew-core/pull/1193. To speed things up here, I also cherry-picked two trivial commits from here into master and just pushed them.

Rebasing and then will try again.

maxim-belkin:
LGTM. Why do you not merge it? The errors are cause by mesa (actually, its patch) so feel free to merge these changes.

Issue #84, opened 2016-05-13
Check build dependencies to make sure they should not be regular dependencies


bw-yuhogg:
Spun off from https://github.com/Linuxbrew/homebrew-xorg/pull/79#issuecomment-218941610

> Since this has bitten us twice now, I think we ought to take a look at the remaining build dependencies and consider changing them to normal dependencies. After all, most users are going to have _almost all_ the formula from this tap installed, assuming they've tapped it at all, by virtue of installing `linuxbrew/xorg/xorg`.

``` bash
$ ag "depends_on.*lib.*build" | grep -v libxslt | grep -v glib | grep -v zlib | grep -v libxml2 | grep -v libtool | grep -v libxi.rb
libxft.rb:21:  depends_on "libxrender" =>  :build
libxrandr.rb:18:  depends_on "libxrender" =>  :build
libdrm.rb:15:  depends_on "libpciaccess" => :build
libxcomposite.rb:19:  depends_on "libxfixes"  =>  :build
```

```
```

bw-yuhogg:
I don't think I'm likely to get around to this any time soon. Unassigning.

bw-yuhogg:
`libxcomposite` only links to `libX11.so`, so there shouldn't be any concerns there. Since the other 3 have been resolved, finally ⌛️  closing this issue

PR #221, opened 2016-11-11
libxft: Depends on libxrender

rwhogg:

No test, so still fails a strict audit.

This is me finally getting around to one of the formulae mentioned on #84.

```bash
/home/linuxbrew/.linuxbrew/Library/Taps/linuxbrew/homebrew-xorg$ patchelf --print-needed /home/linuxbrew/.linuxbrew/opt/libxft/lib/libXft.so
libfontconfig.so.1
libfreetype.so.6
libXrender.so.1
libX11.so.6
libc.so.6
```

### DIFF
--- a/libxft.rb
+++ b/libxft.rb
@@ -15,13 +15,12 @@ class Libxft < Formula
   option "with-brewed-bzip2", "Use brewed bzip2"
   option "with-brewed-zlib", "Use brewed zlib"
 
-  depends_on "pkg-config" =>  :build
-  depends_on "fontconfig" =>  :build
+  depends_on "pkg-config" => :build
 
-  depends_on "libxrender" =>  :build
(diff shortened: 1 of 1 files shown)


libxft.rb
```diff
@@ -15,13 +15,13 @@ class Libxft < Formula
   option "with-brewed-bzip2", "Use brewed bzip2"
   option "with-brewed-zlib", "Use brewed zlib"
 
-  depends_on "pkg-config" =>  :build
-  depends_on "fontconfig" =>  :build
+  depends_on "pkg-config" => :build
+  depends_on "fontconfig" => :build
```
bw-yuhogg: Actually, now that I look at the patchelf output from earlier, it looks like `fontconfig` might be a regular instead of `:build` dependency too.

PR #230, opened 2016-12-17
mesa: Depends on libpthread-stubs (build)

rwhogg:
I'm expecting to find more problems here over time, but mesa has a **long** build time, so if there aren't any, I can just submit this alone.


bw-yuhogg:
`mesa` has other dependency issues to deal with, but I'm not going to worry about them right now.

PR #238, opened 2016-12-28
various: Use HTTPS URL's where available

rwhogg:
Also fixed a few audits here and there


maxim-belkin:
kudos on this one as I've been wanting to switch to https for a while now and kept forgetting.

PR #259, opened 2017-02-11
mesa: Build a bottle for Linuxbrew

maxim-belkin:


sjackman:
Add bottling to `circle.yml`.
Superseded by https://github.com/Linuxbrew/homebrew-xorg/pull/260

PR #269, opened 2017-02-19
xkeyboardconfig: new version 2.20

maxim-belkin:


maxim-belkin:
Incorrect file permissions

maxim-belkin:
rebased on top of the fixed master branch

maxim-belkin:
@sjackman @rwhogg 
This is the first time I think it makes sense to use `=> :run`-time dependency. Here is why:
Configuration step can fail with the following message:
```
configure: WARNING:
                 Required dependencies not found. These dependencies are
                 run-time dependencies only and not required for building.
                 Skip this check with --disable-runtime-deps.

                 Installing this version of xkeyboard-config on a host
                 without the required dependencies may result in unusable
                 keyboard layouts.
                 
configure: error: Package 'pthread-stubs', required by 'xcb', not found
```
configuration step fails if a number of packages are not available. Clearly, we can use `--disable-runtime-deps` but I think it would be better to actually say that those are the run-time and not a build dependencies. thoughts?

xkeyboardconfig.rb
```diff
@@ -1,29 +1,31 @@
...
     sha256 "9487e7667e83f7fb041c35eb8fca4bbfd4c290eadf1c7f514fa57c662171e8b0" => :x86_64_linux
   end
 
   depends_on "pkg-config" => :build
-  depends_on "linuxbrew/xorg/xorg" => :build
   depends_on "intltool" => :build
+  depends_on "linuxbrew/xorg/xproto" => :run
+  depends_on "linuxbrew/xorg/kbproto" => :run
+  depends_on "linuxbrew/xorg/libxcb" => :run
+  depends_on "linuxbrew/xorg/libpthread-stubs" => :run
```
rwhogg: Is `:run` necessary for these? My understanding was we shouldn't have to do this except for stuff like `autoconf` and friends.

rwhogg: Ignore that last comment, I didn't see your comment on the main page.

bw-yuhogg:
I'm not sure what exactly the effect of `:run` is, so I'm unsure, but I do think you make a good case for it.

An alternative might be just to put a comment in the code explaining that these are run-time dependencies.

maxim-belkin:
will have to rebuild a bottle for `libxkbcommon` but, otherwise, success.

sjackman:
I also came to the same conclusion as you, before reading the code. My understanding of `:run` is that it actually means `:build_and_run`, and so is a bit of a misnomer. The `:run` dependency is installed by test-bot both for when building the package and when testing the packag. Its purpose is to silence this particular brew audit warning:
```
  * autoconf dependency should be
      depends_on "autoconf" => :build
    Or if it is indeed a runtime dependency
      depends_on "autoconf" => :run
```
That being the case, I'd be inclined to remove `:run` from these dependencies.

maxim-belkin:
ok, will do

maxim-belkin:
my only question right now is should we add `libpthread-stubs` here or make it a normal dependency for `libxcb`... it is needed only to detect `lixcb`

sjackman:
It sounds as though `libpthread-stubs` should be a normal dependency of `libxcb`. Does the `libxcb.pc` file reference `libpthread-stubs.pc`?

maxim-belkin:
> It sounds as though `libpthread-stubs` should be a normal dependency of `libxcb`. Does the `libxcb.pc` file reference `libpthread-stubs.pc`?

Yes, `xcb.pc` references `pthread-stubs`

maxim-belkin:
I'm not sure why we have to manually list the dependency on `libxcb` even though we depend on `libx11` which depends on `libxcb`. Is this some sort of a bug related to `xorg` tap?

bw-yuhogg:
> I'm not sure why we have to manually list the dependency on libxcb even though we depend on libx11 which depends on libxcb. Is this some sort of a bug related to xorg tap?

Is it still necessary to do that after rebasing on top of #286?

bw-yuhogg:
From Travis (https://travis-ci.org/Linuxbrew/homebrew-xorg/builds/220167652#L3071):

```
Testing libxkbcommon
==> /home/linuxbrew/.linuxbrew/bin/gcc-5 test.c -I/home/linuxbrew/.linuxbrew/Cellar/libxkbcommon/0.7.1/include -L/home/linuxbrew/.linuxbrew/Cellar/libxkbcommon/0.7.1/lib -lxkbcommon -o test
==> ./test
.
Last 150 lines from /home/linuxbrew/.cache/Homebrew/Logs/libxkbcommon/test.02.test:
2017-04-09 02:36:28 +0000
./test
xkbcommon: ERROR: failed to add default include path /home/linuxbrew/.linuxbrew/Cellar/xkeyboardconfig/2.17/share/X11/xkb
Error: libxkbcommon: failed
Failed executing: ./test 
/home/linuxbrew/.linuxbrew/Library/Homebrew/formula.rb:1836:in `block in system'
/home/linuxbrew/.linuxbrew/Library/Homebrew/formula.rb:1774:in `open'
/home/linuxbrew/.linuxbrew/Library/Homebrew/formula.rb:1774:in `system'
/home/linuxbrew/.linuxbrew/Library/Taps/homebrew/homebrew-core/Formula/libxkbcommon.rb:61:in `block in <class:Libxkbcommon>'
/home/linuxbrew/.linuxbrew/Library/Homebrew/formula.rb:1673:in `block (2 levels) in run_test'
/home/linuxbrew/.linuxbrew/Library/Homebrew/formula.rb:884:in `with_logging'
/home/linuxbrew/.linuxbrew/Library/Homebrew/formula.rb:1672:in `block in run_test'
/home/linuxbrew/.linuxbrew/Library/Homebrew/extend/fileutils.rb:14:in `block in mktemp'
/home/linuxbrew/.linuxbrew/Library/Homebrew/extend/fileutils.rb:74:in `block in run'
/home/linuxbrew/.linuxbrew/Library/Homebrew/extend/fileutils.rb:74:in `chdir'
/home/linuxbrew/.linuxbrew/Library/Homebrew/extend/fileutils.rb:74:in `run'
/home/linuxbrew/.linuxbrew/Library/Homebrew/extend/fileutils.rb:13:in `mktemp'
/home/linuxbrew/.linuxbrew/Library/Homebrew/formula.rb:1666:in `run_test'
/home/linuxbrew/.linuxbrew/Library/Homebrew/test.rb:33:in `block in <main>'
/home/travis/.rvm/rubies/ruby-2.0.0-p648-clang/lib/ruby/2.0.0/timeout.rb:66:in `timeout'
/home/linuxbrew/.linuxbrew/Library/Homebrew/test.rb:32:in `<main>'
```

maxim-belkin:
we'll have to rebuild `libxkbcommon` after this PR.

PR #270, opened 2017-02-19
libxfont2: new formula

maxim-belkin:


maxim-belkin:
replacement for `libxfont`

maxim-belkin:
turns out, `libxfont` is needed by `bdftopcf`... so, instead of replacing `libxfont`, add `libxfont2` instead.

PR #274, opened 2017-02-20
libxfont: fixes + test block

maxim-belkin:


maxim-belkin:
`libxfont2` provides a library with a different name, so there is no need to remove `libxfont`. Instead, let's fix `libxfont`

PR #277, opened 2017-03-07
mesa: change dependency on xorg to :x11

maxim-belkin:


ilovezfs:
@maxim-belkin https://ftp.cc.uoc.gr/mirrors/linux/lfs/LFS/conglomeration/Xorg/mesa-13.0.4-add_xdemos-1.patch

maxim-belkin:
@sjackman shall we merge this one?

rwhogg (approved):

maxim-belkin:
Thanks, Bob! we can skip updating the bottle, right?

bw-yuhogg:
They're just demos, right? If so, yeah, IMO there's no real need to update the bottle.

sjackman (approved):

sjackman:
:shipit:

PR #279, opened 2017-03-09
prepend linuxbrew/xorg to dependencies from linuxbrew/xorg tap

maxim-belkin:
@sjackman, please review

Used script:
```sh
for file in *.rb
do
  name=${file%%.rb}
  echo $name
  for rfile in $(grep -l "depends_on\s\+\"$name\"" *)
  do
    sed -i -e "s|depends_on\(\s\+\)\"$name\"|depends_on\1\"linuxbrew/xorg/$name\"|" "$rfile"
  done
done
```


sjackman (approved):

sjackman:
Looks good to me.

Issue #280, opened 2017-04-04
libx11: Requested 'xcb >= 1.11.1' but version of XCB is 1.11


sjackman:
_From @parsibox on April 4, 2017 7:50_

hi
can not install X11

```
[parsibox@localhost ~]$ brew install linuxbrew/xorg/libx11
==> Installing libx11 from linuxbrew/xorg
==> Downloading http://ftp.x.org/pub/individual/lib/libX11-1.6.4.tar.bz2
Already downloaded: /home/parsibox/.cache/Homebrew/libx11-1.6.4.tar.bz2
==> Downloading https://raw.githubusercontent.com/Linuxbrew/homebrew-xorg/master/patch_configure.diff
Already downloaded: /home/parsibox/.cache/Homebrew/libx11--patch-e3aff4be9c8a992fbcbd73fa9ea6202691dd0647f73d1974ace537f3795ba15f.diff
==> Patching
==> Applying patch_configure.diff
patching file configure
Hunk #1 succeeded at 18432 (offset -215 lines).
==> ./configure --prefix=/home/parsibox/.linuxbrew/Cellar/libx11/1.6.4 --sysconfdir=/home/parsibox/.linuxbrew/etc --localstatedir=/home/parsibox/.linuxbrew/var --disable-silent-rules --enable-static=no --enable-specs=no
Last 15 lines from /home/parsibox/.cache/Homebrew/Logs/libx11/01.configure:
checking for sys/param.h... yes
checking for getpagesize... yes
checking for working mmap... yes
checking for nl_langinfo... yes
checking for X11... no
configure: error: Package requirements (xproto >= 7.0.17 xextproto xtrans xcb >= 1.11.1 kbproto inputproto) were not met:

Requested 'xcb >= 1.11.1' but version of XCB is 1.11

Consider adjusting the PKG_CONFIG_PATH environment variable if you
installed software in a non-standard prefix.

Alternatively, you may set the environment variables X11_CFLAGS
and X11_LIBS to avoid the need to call pkg-config.
See the pkg-config man page for more details.



```

```
[parsibox@localhost ~]$ brew config
HOMEBREW_VERSION: 1.1.10
ORIGIN: https://github.com/Linuxbrew/brew.git
HEAD: d8c8b867bf99c604c89bc848f58d1fae8afd599c
Last commit: 6 weeks ago
Core tap ORIGIN: https://github.com/Linuxbrew/homebrew-core
[...]

PR #281, opened 2017-04-04
xcb-proto: new version 1.12

maxim-belkin:


maxim-belkin:
Step 1/2 in updating `libxcb`.
Related to #280.

PR #285, opened 2017-04-07
xkeyboardconfig: change the dependency on xorg to :x11

maxim-belkin:


maxim-belkin:
Actually, closing this one in favor of #269

PR #286, opened 2017-04-08
libxcb: make libpthread-stubs a normal dependency

maxim-belkin:


maxim-belkin:
Related to Linuxbrew/homebrew-core#2302
Should make `libpthread-stubs` not necessary for formula that depend on `xcb`

rwhogg (approved):
I think a comment explaining that the xcb.pc file references pthread-stubs would be helpful, but it's not necessary IMO.

Note: I did a mass installation of xorg stuff today. Excluding stuff that failed to build, no other pkg-config files reference pthread-stubs, so I don't think this change is needed anywhere else.

maxim-belkin:
Thanks, Bob! I've added a comment and merged the PR without updating the bottle.

PR #290, opened 2017-04-09
TEST: see what happens when xkeyboardconfig depends strictly on libx11

rwhogg:
Do not merge this; it's just a test to see what happens when `xkeyboardconfig` depends solely on `libx11`. (Part of reviewing #269)


maxim-belkin:
nice! so it looks like `libx11` brings everything `xkeyboardconfig` needs!

bw-yuhogg:
Closing as this is no longer needed.

PR #289, opened 2017-04-09
libvdpau-va-gl: depends on glib for Linuxbrew

rwhogg:
I'm expecting this to fail; I just want to see *how* it fails.

PR #291, opened 2017-04-09
libva: build X11 bindings

rwhogg:
autoreconf is only called if wayland is not requested,
but if it is, libtool is needed.

Signed-off-by: Bob W. Hogg <rwhogg@linux.com>

- [ ] Have you included a log of the failed build without your patch using `brew gist-logs <formula>`?

### DIFF
--- a/libva.rb
+++ b/libva.rb
@@ -26,6 +26,10 @@ class Libva < Formula
   depends_on "linuxbrew/xorg/libdrm"
   depends_on "linuxbrew/xorg/wayland" => :recommended
 
+  # Needed for X11 bindings
+  depends_on "linuxbrew/xorg/libxext"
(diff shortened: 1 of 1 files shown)


bw-yuhogg:
We need to make sure that the bottle for `libva` includes libva-x11.so and libva-x11.pc, because without them `libvdpau-va-gl` does not build (#289). It might be possible to change that formula to work with Wayland, but I'm not sure how and I don't know much about Wayland.

It seems that `libva` includes the X11 bindings if built in stdenv (on my machine), but not if built in superenv. I'll have to work out what dependencies we need for X11.

Thoughts @maxim-belkin?

bw-yuhogg:
I'm not sure if this is still necessary, but the `libtool` dependency is certainly still necessary. I'll close this and reopen one with just the `libtool` change.

PR #296, opened 2017-05-01
CI: make Travis & Circle consistent with core

maxim-belkin:


.travis.yml
```diff
@@ -37,19 +42,22 @@ before_install:
   - sudo mkdir $HOME
   - sudo chown "$USER:" $HOME
   - LINUXBREW=$HOME/.linuxbrew
-  - git clone https://github.com/Linuxbrew/brew $LINUXBREW
+  - git clone --depth=1 https://github.com/Linuxbrew/brew "$LINUXBREW"
```
bw-yuhogg, line 45: Core currently has ".git" at the end of the URL; while I had previously removed that from core (https://github.com/Linuxbrew/homebrew-core/pull/1790), both core and science currently appear to have it, so it might make sense to put it back in and reduce the diff size between repos.

sjackman: Let's remove `.git` from Homebrew/science.
*Important note*. `--depth=1` makes the output of `brew --version` less informative, and affects info stored in the bottle. So it's fine for `.travis.yml` but no good for `circle.yml` where we build bottles.

circle.yml
```diff
@@ -25,16 +25,20 @@ dependencies:
     - DEBIAN_FRONTEND=noninteractive sudo apt-get autoremove -y --purge
         libblas-dev
         libbz2-dev
+        libexpat1-dev
```
bw-yuhogg, line 28: Does doing this remove `python-dev`?

https://github.com/Homebrew/brew/releases/tag/1.1.12 changes some behavior around when the `default_formula` is installed, so what we want to do about `python-dev` and `libexpat1-dev` might change after we merge 1.1.12 and later.

I'm not too familiar with the changes; see https://github.com/Homebrew/brew/pull/2389

sjackman (approved):

.travis.yml
maxim-belkin: So, should I remove `.git` from here then? 🙃

circle.yml
maxim-belkin: > Does doing this remove `python-dev`?

I think so. See https://linuxbrew.slack.com/archives/C0R1UM07K/p1493658027563209

sjackman: Hmm. Currently we intentionally install `python2.7-dev`. A reasonable option would be to use the brewed Python for libraries and headers. So options are…

1. Install `python2.7` and `python2.7-dev`
2. Install `python2.7` but not `python2.7-dev`
3. Install neither `python2.7` nor `python2.7-dev`

I don't know whether the the `:python` requirement currently checks that the Python header/libraries are installed. That'd be needed for option 2. Currently we use option 1.

bw-yuhogg: > I don't know whether the the :python requirement currently checks that the Python header/libraries are installed. That'd be needed for option 2.

I quickly glanced at it; looks like it doesn't.

I'm approving the changes now; we can postpone this discussion until after 1.1.12 is merged.

rwhogg (approved):
Thanks @maxim-belkin

.travis.yml
sjackman: Yes, I think so.

.travis.yml
```diff
@@ -1,23 +1,29 @@
...
   - DEBIAN_FRONTEND=noninteractive sudo apt-get autoremove -y --purge
       libblas-dev
       libbz2-dev
+      libexpat1-dev
       libfreetype6-dev
       libgdbm-dev
       libicu-dev
       libidn11-dev
+      libjack-jackd2-0
+      libjack0
```
sjackman, line 19: For consistency, remove the `-dev` packages rather than the libraries.

maxim-belkin: I can't find corresponding `-dev` packages. Which ones do you mean?

maxim-belkin: `libjack-dev`?

maxim-belkin: and also maybe `libjack-jackd2-dev`

sjackman: Yep, sounds good.

PR #306, opened 2017-06-15
Mass style fix: remove extraneous spaces after url

rwhogg:
Signed-off-by: Bob W. Hogg <rwhogg@linux.com>

PR #307, opened 2017-06-16
Mass audit fix: use %i for arrays of symbols

rwhogg:
Signed-off-by: Bob W. Hogg <rwhogg@linux.com>

For clarity:

```ruby
%i(build recommended) == [:build, :recommended]
# => true
```

These were failing audits left and right.

Note that I still have yet to fix the HTTPS URL's for a lot of these formulae; I'll try to do that after #306 is merged, because it would introduce a lot of merge conflicts, which I'd prefer to avoid.

### DIFF
--- a/bigreqsproto.rb
+++ b/bigreqsproto.rb
@@ -23,8 +23,8 @@ class Bigreqsproto < Formula
 
   if build.with?("specs")
     depends_on "xmlto" => :build
-    depends_on "fop" => [:build, :recommended]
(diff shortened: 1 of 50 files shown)


maxim-belkin:
why do you change it to `%i(...)` ?

bw-yuhogg:
> why do you change it to `%i(...)` ?

Huh, that's weird. Previously they were failing audits with the old syntax. Now they aren't, but *this branch* is failing audits. Guess something changed in `brew` recently.

I'll close this.

PR #329, opened 2017-08-29
libva: new version 1.8.2 and reordered dependencies

maxim-belkin:
Also added dependencies that `brew linkage` reported as possibly missing:

-  linuxbrew/xorg/libx11
-  linuxbrew/xorg/libxext
-  linuxbrew/xorg/libxfixes

(even though installation succeeds without them being initially installed)

### DIFF
--- a/Formula/libva.rb
+++ b/Formula/libva.rb
@@ -15,7 +15,6 @@ class Libva < Formula
 
   option "with-eglx", "Build libva with egl and glx support (use after building mesa)"
 
-  # Build-time
   depends_on "autoconf" => :build
   depends_on "pkg-config" => :build
   depends_on "libtool" => :build if build.without?("wayland")
--- a/Formula/libva.rb
+++ b/Formula/libva.rb
@@ -1,12 +1,8 @@
 class Libva < Formula
   desc "Hardware accelerated video processing library"
   homepage "https://freedesktop.org/wiki/Software/vaapi/"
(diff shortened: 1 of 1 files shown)


libva.rb
```diff
@@ -20,14 +16,15 @@ class Libva < Formula
...
   # Build-time
-  depends_on "pkg-config" => :build
   depends_on "autoconf" => :build
-
+  depends_on "pkg-config" => :build
+  depends_on "libtool" => :build if build.without?("wayland")
   depends_on "linuxbrew/xorg/libdrm"
+  depends_on "linuxbrew/xorg/libx11"
+  depends_on "linuxbrew/xorg/libxext"
+  depends_on "linuxbrew/xorg/libxfixes"
```
bw-yuhogg: My understanding is that `libva` will build without X11 if those libraries aren't present.

Since it isn't a part of core X11, might it make sense for `libva` to just depend on `:x11` rather than specific libraries? We can also make it `:recommended` if you think that'd make more sense; I'm no expert on it.

maxim-belkin:
Currently blocked by a bug in `brew linkage` logic

maxim-belkin:
Bug should be fixed in https://github.com/Linuxbrew/brew/pull/634

maxim-belkin:
Should be merged together with mesa (no corresponding PR yet)

maxim-belkin:
Libva was updated in a different PR

PR #334, opened 2017-10-26
wayland-protocols 1.11

iMichka:


iMichka:
Will need #333 first.

PR #347, opened 2018-01-30
mesa: Use xorg dep instead of X11 for Linuxbrew

iMichka:


mesa.rb
```diff
@@ -3,10 +3,9 @@ class Mesa < Formula
   homepage "https://dri.freedesktop.org"
   url "https://mesa.freedesktop.org/archive/mesa-17.2.3.tar.xz"
   sha256 "a0b0ec8f7b24dd044d7ab30a8c7e6d3767521e245f88d4ed5dd93315dc56f837"
-  revision 1
+  revision 2
 
   bottle do
-    sha256 "3642e9481487eb1acf28f6a4a3ea8af4b872a6c9b3f55426790bc016aae01599" => :x86_64_linux
```
sjackman, line 9: There's no need to remove the bottle line.

iMichka: I will change this locally with a rebase, because I do not have 2 hours to wait for it to build again :)

PR #349, opened 2018-02-13
Moving formulas to Formula folder

maxim-belkin:
--
Moving formulas to "Formula" subfolder.


Formula/bigreqsproto.rb
```diff
@@ -17,7 +17,7 @@ class Bigreqsproto < Formula
 
   # Patch for xmlto
   patch do
-    url "https://raw.githubusercontent.com/Linuxbrew/homebrew-xorg/master/patch_configure.diff"
+    url "https://raw.githubusercontent.com/Linuxbrew/homebrew-xorg/master/Formula/patch_configure.diff"
```
sjackman: Patches could go in their own subdirectory, `Patches`. I usually prefer to `gist` them myself.

sjackman (changes requested):
You'll need to update the CircleCI and Travis configurations.

Formula/bigreqsproto.rb
maxim-belkin: I like the idea of `Patches` folder :)
Gists are fine too though I think keeping patches affiliated with the project is a better approach.

sjackman: Either is fine by me. Be sure to link to an absolute revision of the patch and not `master`.

maxim-belkin: okeydoke

maxim-belkin: I guess the patch has be to moved to Patches directory first to record the commit number

sjackman: Good point. That may be why upstream uses a separate repo to store patches, and why I use gists. (Never thought about it before, myself.)

.circleci/config.yml
```diff
@@ -20,10 +20,12 @@ jobs:
...
       - run: git config --global user.name LinuxbrewTestBot
       - run: git config --global user.email testbot@linuxbrew.sh
-      - run: chmod 0644 *.rb
+      - run: chmod 0644 Formula/*.rb
       - run: mkdir -p /home/linuxbrew/.linuxbrew/Homebrew/Library/Taps/linuxbrew
       - run: cp -a . /home/linuxbrew/.linuxbrew/Homebrew/Library/Taps/linuxbrew/$CIRCLE_PROJECT_REPONAME
-      - run: |
+      - run:
+         no_output_timeout: 30m
+         command: |
```
sjackman: The indentation here is odd. There's three spaces of indent between run and command, and only one space between command and the next chunk.

.travis.yml
```diff
@@ -82,23 +81,29 @@ before_install:
...
-  - chmod 0644 *.rb
+  - chmod 0644 Formula/*.rb
   - umask 022
-  - export HOMEBREW_DEVELOPER=1
-  - export HOMEBREW_NO_AUTO_UPDATE=1
-  - export HOMEBREW_VERBOSE=1 HOMEBREW_VERBOSE_USING_DOTS=1
   - env | grep TRAVIS | tee /tmp/travis.env
   - ulimit -n 1024
-  - brew tap homebrew/science
+  - brew tap linuxbrew/extra
```
sjackman: I'd suggest removing this line if this tap is not used by Linuxbrew/org.

.travis.yml
```diff
@@ -82,23 +81,29 @@ before_install:
...
+matrix:
+  fast_finish: true
+
 script:
-  - brew test-bot
+  - brew test-bot --bintray-org=linuxbrew --git-name=LinuxbrewTestBot --git-email=testbot@linuxbrew.sh
 
 notifications:
+  slack:
+    secure: PvK4LXijxgFTbT6SvWUveluJdbe7AfnJ+Ip/xb2gVRsq71DHSi96OtaEZq+qAvU6UU07VaCY/55cTb+HZE1twCOE4aE4l6iJONVmwIb+s9bGQfV9+H/r+y+J1EJzIzAfTbsn4mGJpkG9e2326c0vGcM/lFpHH0qarhOEYs4jLt17Ijg4mk+I5jLIOZ2vZsC7IZ0W8oe7Qzomyyfzj5XW2kUa41e8CgZf4byqAvnVpPpfvkm/rODGPREOmcpOr2Pu8fhNLrswXRlhgakZgi5JPPdaRocIhnbRbyqE2Qngnlk/wzno1e3NZYDcxTa5xAgAa5h/TIpyLeB/pXfpc1n5RuP7BjA/mZSYaG5Dg1IIvGPu/23+HlZ9umMn0E0f5j3EsV66CwErJGChm6hN7xPAxLz1Mj9v2E3CIDjyYcg/aZ+O27oaUlGoBGrX51g8poq6LhG+SxHx/VkDAJ+X4T56EQYTCA+ezEIs6KdWZl7AB4ztHqTe7w4YJIIhh509iViFKX4pJdQnajNf97IvRl9hHsLa9k9GuLWOQvxjkEJ/1LSfVkNEfPxHd2btIVdg6DQsVrxyfwufF+KlQDKNHR4Jbz+KEvrWIbXWQAj+2g7xwj5l2/Tyv69gR7FqBpSxh+Ue7wrWrVCIkzn/BEU4H0nHGR5D4dOwmgiOxXVhEZjYqjU=
```
sjackman: Please remove these two lines. They are specific to each repo (I believe).

sjackman (changes requested):

.circleci/config.yml
```diff
@@ -4,37 +4,48 @@ jobs:
...
-      - run: |
+      - run:
+        name: Set up Git
+        command: |
+          git config --global user.name LinuxbrewTestBot
+          git config --global user.email testbot@linuxbrew.sh
+      - run:
+        name: Set up Linuxbrew
+        working_directory: /home/linuxbrew/.linuxbrew/Homebrew
+        no_output_timeout: 5m
```
sjackman: The default is 10 minutes. Why are you decreasing it to 5 minutes? I'd leave it at the default or increase it.

.circleci/config.yml
```diff
@@ -4,37 +4,48 @@ jobs:
...
+        command: |
+          git pull --ff-only
+      - run:
+        name: Set up Linuxbrew/$CIRCLE_PROJECT_REPONAME
+        no_output_timeout: 5m
```
sjackman: Ditto

.circleci/config.yml
sjackman: You can delete this line if you're happy with the default value.

maxim-belkin:
Hmmmm...
CI gives an error:

> * In step 1 definition: Invalid step structure (expected string or map, got config.StepDescription)

.circleci/config.yml
```diff
@@ -4,37 +4,48 @@ jobs:
...
+        name: Set up Git
+        command: |
+          git config --global user.name LinuxbrewTestBot
+          git config --global user.email testbot@linuxbrew.sh
+      - run:
+        name: Set up Linuxbrew
+        working_directory: /home/linuxbrew/.linuxbrew/Homebrew
+        no_output_timeout: 5m
+        command: |
+          git pull --ff-only
```
sjackman: You shouldn't need this section. The `linuxbrew/linuxbrew` Docker image tracks `master` of `Linuxbrew/brew`.

maxim-belkin: true :)

.circleci/config.yml
```diff
@@ -4,37 +4,48 @@ jobs:
...
+      - run:
+        name: Set up Linuxbrew/$CIRCLE_PROJECT_REPONAME
+        no_output_timeout: 5m
+        command: |
+          if [[ -n "$CIRCLE_PR_NUMBER" ]]; then
+              git fetch https://github.com/Linuxbrew/$CIRCLE_PROJECT_REPONAME pull/$CIRCLE_PR_NUMBER/head
```
sjackman: I suggest `git fetch origin pull/$CIRCLE_PR_NUMBER/head`

.circleci/config.yml
```diff
@@ -4,37 +4,48 @@ jobs:
...
+          fi
+          git reset --hard $CIRCLE_SHA1
+          mkdir -p /home/linuxbrew/.linuxbrew/Homebrew/Library/Taps/linuxbrew
+          cp -a . /home/linuxbrew/.linuxbrew/Homebrew/Library/Taps/linuxbrew/$CIRCLE_PROJECT_REPONAME
```
sjackman: Skip the copy and clone the repo to the directory you want it in. I don't see a `git clone`. What's in the current working directory?

.circleci/config.yml
maxim-belkin: 10 minutes might be an overkill for this step, don't you think?

.circleci/config.yml
maxim-belkin: ok

.circleci/config.yml
sjackman: Sure, but I'd rather not clutter up the config with items that won't make a difference.

maxim-belkin: ok, will remove

.circleci/config.yml
```diff
@@ -15,24 +15,21 @@ jobs:
...
-              git fetch https://github.com/Linuxbrew/$CIRCLE_PROJECT_REPONAME pull/$CIRCLE_PR_NUMBER/head
+              git fetch origin pull/$CIRCLE_PR_NUMBER/head
           fi
           git reset --hard $CIRCLE_SHA1
-          mkdir -p /home/linuxbrew/.linuxbrew/Homebrew/Library/Taps/linuxbrew
-          cp -a . /home/linuxbrew/.linuxbrew/Homebrew/Library/Taps/linuxbrew/$CIRCLE_PROJECT_REPONAME
       - run:
         name: Prepare for the main test
-        no_output_timeout: 5m
+        no_output_timeout: 5m
```
sjackman: Remove this line

.circleci/config.yml
```diff
@@ -15,24 +15,21 @@ jobs:
...
-        command: |
-          git pull --ff-only
+        name: Create a directory for Linux taps
+        command: mkdir -p /home/linuxbrew/.linuxbrew/Homebrew/Library/Taps/linuxbrew
       - run:
         name: Set up Linuxbrew/$CIRCLE_PROJECT_REPONAME
-        no_output_timeout: 5m
+        working_directory: /home/linuxbrew/.linuxbrew/Homebrew/Library/Taps/linuxbrew
         command: |
+          git clone $CIRCLE_REPOSITORY_URL $CIRCLE_PROJECT_REPONAME
```
sjackman: Note that `$CIRCLE_REPOSITORY_URL` uses ssh. I'd suggest using
```
git clone https://github.com/$CIRCLE_PROJECT_USERNAME/$CIRCLE_PROJECT_REPONAME
```

sjackman: The second argument `$CIRCLE_PROJECT_REPONAME` of `git clone` is redundant, since it's the default.

maxim-belkin: Hmmm, according to https://circleci.com/docs/1.0/environment-variables/, it's `https://...`

maxim-belkin: oh, it might be 1.0-specific

maxim-belkin:
```
* In step 1 definition: Step

  - run: 
    command: git config --global user.name LinuxbrewTestBot
git config --global user.email testbot@linuxbrew.sh

    name: Set up Git

has incorrect indentation, it should be formatted like:

  - step:
      option1: value
      option2: value
```

sjackman:
Try…
```yaml
  - run:
      name: Set up Git
      command: |
        git config --global user.name LinuxbrewTestBot
        git config --global user.email testbot@linuxbrew.sh
```

maxim-belkin:
yes, that's what I'm changing to now :)

maxim-belkin:
What's interesting is that I ran it through `circleci config validate --verbose -c config.yml` on a mac it  reported that `config file is valid`

maxim-belkin:
Looks like the config is working... but I'm not sure why `brew readall --aliases linuxbrew/xorg` is taking so long... On my local machine it takes  0.5 sec

maxim-belkin:
Expected failure
```
==> Downloading https://raw.githubusercontent.com/Linuxbrew/homebrew-xorg/master/Patches/patch_configure.diff
/usr/bin/curl --show-error --user-agent Linuxbrew/1.X.Y (Linux; x86_64 4.4.0-116-generic) curl/7.47.0 --fail --silent --location --remote-time --continue-at - --output /home/linuxbrew/.cache/Homebrew/xproto--patch-e3aff4be9c8a992fbcbd73fa9ea6202691dd0647f73d1974ace537f3795ba15f.diff.incomplete https://raw.githubusercontent.com/Linuxbrew/homebrew-xorg/master/Patches/patch_configure.diff
curl: (22) The requested URL returned error: 404 Not Found
```

.circleci/config.yml
```diff
@@ -4,32 +4,39 @@ jobs:
...
+      - run:
+          name: Create a directory for Linux taps
+          command: mkdir -p /home/linuxbrew/.linuxbrew/Homebrew/Library/Taps/linuxbrew
+      - run:
+          name: Set up linuxbrew/xorg tap
+          working_directory: /home/linuxbrew/.linuxbrew/Homebrew/Library/Taps/linuxbrew
+          command: |
+            git clone https://github.com/$CIRCLE_PROJECT_USERNAME/$CIRCLE_PROJECT_REPONAME
+            cd $CIRCLE_PROJECT_REPONAME
+            if [[ -n "$CIRCLE_PR_NUMBER" ]]; then
```
maxim-belkin: I wonder if we should test for this early and execute the test ONLY if CIRCLE_PR_NUMBER is present (that is, don't run circleci for direct commits... which should not happen, but still...)

maxim-belkin:
I'm proposing that we have different tests for different PRs:

- for PRs like this one when we don't need to build new bottles
- for PRs that build bottles
- for direct commits (again, we don't need to build bottles for these and it's "too late").

maxim-belkin:
The above comments should not affect this PR though as we can implement all of them later

.circleci/config.yml
sjackman: `test-bot` is run for commits to `master`. It checks syntax and style.

sjackman:
> I'm proposing that we have different tests for different PRs.

We currently have two different tests: tests for pull requests, which build bottles, and tests for commits to master. Pull requests that don't build bottles are less common and may not warrant a special case.

.circleci/config.yml
maxim-belkin: Apologies, but I'm not sure what you mean... `test-bot` ("Main test") is executed for PRs too, including this one: https://circleci.com/gh/Linuxbrew/homebrew-xorg/977

maxim-belkin:
> We currently have two different tests: tests for pull requests, which build bottles, and tests for commits to master.

I don't see where we define different tests for commits and PRs

sjackman:
https://github.com/Linuxbrew/homebrew-test-bot/blob/30e592d0bc665bccbed0df66902ce2eb228f9700/cmd/brew-test-bot.rb#L429

https://github.com/Linuxbrew/homebrew-test-bot/blob/30e592d0bc665bccbed0df66902ce2eb228f9700/cmd/brew-test-bot.rb#L489

.circleci/config.yml
sjackman: `test-bot` is run for both commits to `master` and for PRs. It has a separate mode of operation for each. That latter builds bottles. The former only checks syntax.

sjackman (approved):

sjackman:
It'd be best I think to separate this PR into moving the formulae to `Formula/` with minimal changes to `.circleci/config.yml`. Then a separate PR for your rework of the CircleCI config. That second PR should be 🍏.

.circleci/config.yml
```diff
@@ -4,32 +4,39 @@ jobs:
...
-          brew install patchelf
-          brew test-bot --tap=linuxbrew/xorg --bintray-org=linuxbrew --git-name=LinuxbrewTestBot --git-email=testbot@linuxbrew.sh
+      - run:
+          name: Set up Git
+          command: |
+            git config --global user.name LinuxbrewTestBot
+            git config --global user.email testbot@linuxbrew.sh
+      - run:
+          name: Create a directory for Linux taps
+          command: mkdir -p /home/linuxbrew/.linuxbrew/Homebrew/Library/Taps/linuxbrew
```
sjackman: You can remove this chunk. `git` will make this directory for you.

maxim-belkin: > `git` will make this directory for you.

How exactly is it going to create one?

sjackman: One directory at a time. :wink:

maxim-belkin:
Reverted the changes to CIRCLECI that are not needed in this PR.  The only real fix that we need is removing `chmod *.rb` because `linuxbrew/linuxbrew` has everything working correctly already.

PR #350, opened 2018-03-02
CircleCI changes

maxim-belkin:
--

Changes to CircleCI that were originally developed in #349

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,10 +16,10 @@ jobs:
             git config --global user.email testbot@linuxbrew.sh
       - run:
           name: Set up linuxbrew/xorg tap
-          working_directory: /home/linuxbrew/.linuxbrew/Homebrew/Library/Taps
+          working_directory: /home/linuxbrew/.linuxbrew/Homebrew/Library/Taps/linuxbrew
           command: |
-            git clone https://github.com/$CIRCLE_PROJECT_USERNAME/$CIRCLE_PROJECT_REPONAME linuxbrew/$CIRCLE_PROJECT_REPONAME
-            cd linuxbrew/$CIRCLE_PROJECT_REPONAME
+            git clone https://github.com/$CIRCLE_PROJECT_USERNAME/$CIRCLE_PROJECT_REPONAME
+            cd $CIRCLE_PROJECT_REPONAME
             if [[ -n "$CIRCLE_PR_NUMBER" ]]; then
                 git fetch origin pull/$CIRCLE_PR_NUMBER/head
             fi
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,7 +18,7 @@ jobs:
           name: Set up linuxbrew/xorg tap
           working_directory: /home/linuxbrew/.linuxbrew/Homebrew/Library/Taps/linuxbrew
           command: |
-            git clone https://github.com/$CIRCLE_PROJECT_USERNAME/$CIRCLE_PROJECT_REPONAME
+            git clone https://github.com/Linuxbrew/$CIRCLE_PROJECT_REPONAME
             cd $CIRCLE_PROJECT_REPONAME
             if [[ -n "$CIRCLE_PR_NUMBER" ]]; then
                 git fetch origin pull/$CIRCLE_PR_NUMBER/head
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,31 +4,36 @@ jobs:
     docker:
       - image: linuxbrew/linuxbrew
     environment:
-      CIRCLE_REPOSITORY_URL: https://github.com/linuxbrew/homebrew-xorg
       HOMEBREW_DEVELOPER: 1
       HOMEBREW_NO_AUTO_UPDATE: 1
       HOMEBREW_VERBOSE: 1
       HOMEBREW_VERBOSE_USING_DOTS: 1
     steps:
-      - run: |
-          cd /home/linuxbrew/.linuxbrew/Homebrew
-          git fetch origin --unshallow --tags
-          git reset --hard origin/master
-      - checkout
-      - run: git remote set-url origin $CIRCLE_REPOSITORY_URL
-      - run: if [ -e .git/shallow ]; then echo git fetch --unshallow; fi
-      - run: git fetch origin
-      - run: git config --global user.name LinuxbrewTestBot
-      - run: git config --global user.email testbot@linuxbrew.sh
-      - run: mkdir -p /home/linuxbrew/.linuxbrew/Homebrew/Library/Taps/linuxbrew
-      - run: cp -a . /home/linuxbrew/.linuxbrew/Homebrew/Library/Taps/linuxbrew/$CIRCLE_PROJECT_REPONAME
-      - run: |
-          mkdir /tmp/bottles
-          cd /tmp/bottles
-          umask 022
-          PATH="/home/linuxbrew/.linuxbrew/bin:/home/linuxbrew/.linuxbrew/sbin:$PATH"
-          brew install patchelf
-          brew test-bot --tap=linuxbrew/xorg --bintray-org=linuxbrew --git-name=LinuxbrewTestBot --git-email=testbot@linuxbrew.sh
+      - run:
+          name: Set up Git
+          command: |
+            git config --global user.name LinuxbrewTestBot
+            git config --global user.email testbot@linuxbrew.sh
+      - run:
+          name: Set up linuxbrew/xorg tap
+          working_directory: /home/linuxbrew/.linuxbrew/Homebrew/Library/Taps
+          command: |
+            git clone https://github.com/$CIRCLE_PROJECT_USERNAME/$CIRCLE_PROJECT_REPONAME linuxbrew/$CIRCLE_PROJECT_REPONAME
+            cd linuxbrew/$CIRCLE_PROJECT_REPONAME
+            if [[ -n "$CIRCLE_PR_NUMBER" ]]; then
+                git fetch origin pull/$CIRCLE_PR_NUMBER/head
+            fi
+            git reset --hard $CIRCLE_SHA1
+      - run:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -4,31 +4,36 @@ jobs:
...
-      - run: cp -a . /home/linuxbrew/.linuxbrew/Homebrew/Library/Taps/linuxbrew/$CIRCLE_PROJECT_REPONAME
-      - run: |
-          mkdir /tmp/bottles
-          cd /tmp/bottles
-          umask 022
-          PATH="/home/linuxbrew/.linuxbrew/bin:/home/linuxbrew/.linuxbrew/sbin:$PATH"
-          brew install patchelf
-          brew test-bot --tap=linuxbrew/xorg --bintray-org=linuxbrew --git-name=LinuxbrewTestBot --git-email=testbot@linuxbrew.sh
+      - run:
+          name: Set up Git
```
sjackman, line 13: My personal preference is to omit `name` for conciseness. Feel free to do as you prefer in this tap.

.circleci/config.yml
```diff
@@ -4,31 +4,36 @@ jobs:
...
+          command: |
+            git config --global user.name LinuxbrewTestBot
+            git config --global user.email testbot@linuxbrew.sh
+      - run:
+          name: Set up linuxbrew/xorg tap
+          working_directory: /home/linuxbrew/.linuxbrew/Homebrew/Library/Taps
```
sjackman: I'd suggest removing this line.

maxim-belkin: that would entail changing `git clone` and `cd` commands below to be very long

.circleci/config.yml
```diff
@@ -4,31 +4,36 @@ jobs:
...
+          command: |
+            git clone https://github.com/$CIRCLE_PROJECT_USERNAME/$CIRCLE_PROJECT_REPONAME linuxbrew/$CIRCLE_PROJECT_REPONAME
```
sjackman: I suggest…
```
mkdir -p /home/linuxbrew/.linuxbrew/Homebrew/Library/Taps/$CIRCLE_PROJECT_USERNAME
cd /home/linuxbrew/.linuxbrew/Homebrew/Library/Taps/$CIRCLE_PROJECT_USERNAME
git clone https://github.com/$CIRCLE_PROJECT_USERNAME/$CIRCLE_PROJECT_REPONAME
cd $CIRCLE_PROJECT_REPONAME
```
or
```
git clone https://github.com/$CIRCLE_PROJECT_USERNAME/$CIRCLE_PROJECT_REPONAME /home/linuxbrew/.linuxbrew/Homebrew/Library/Taps/$CIRCLE_PROJECT_USERNAME/$CIRCLE_PROJECT_REPONAME
cd /home/linuxbrew/.linuxbrew/Homebrew/Library/Taps/$CIRCLE_PROJECT_USERNAME/$CIRCLE_PROJECT_REPONAME
```

.circleci/config.yml
sjackman: On second thought, I think I like yours better.
If the directory specified by `working_directory` doesn't exist, what does CircleCI do? Does it create it, or give an error?

maxim-belkin: Interestingly enough, I think it is created if it does not exist
See
https://circleci.com/docs/2.0/configuration-reference/#job_name
below
https://circleci.com/docs/2.0/configuration-reference/#parallelism

It defaults to.... :man_facepalming: ... ~/projects

.circleci/config.yml
```diff
@@ -4,31 +4,36 @@ jobs:
...
+      - run:
+          name: Set up Git
+          command: |
+            git config --global user.name LinuxbrewTestBot
+            git config --global user.email testbot@linuxbrew.sh
+      - run:
+          name: Set up linuxbrew/xorg tap
+          working_directory: /home/linuxbrew/.linuxbrew/Homebrew/Library/Taps/linuxbrew
+          command: |
+            git clone https://github.com/$CIRCLE_PROJECT_USERNAME/$CIRCLE_PROJECT_REPONAME
```
maxim-belkin: Not sure we should keep `$CIRCLE_PROJECT_USERNAME` or replace it with a shorter and clearer `Linuxbrew`

maxim-belkin:
```
  CI=true
  CIRCLECI=true
  CIRCLE_BRANCH=pull/350
  CIRCLE_BUILD_NUM=981
  CIRCLE_BUILD_URL=https://circleci.com/gh/Linuxbrew/homebrew-xorg/981
  CIRCLE_COMPARE_URL=
  CIRCLE_JOB=build
  CIRCLE_NODE_INDEX=0
  CIRCLE_NODE_TOTAL=1
  CIRCLE_PREVIOUS_BUILD_NUM=980
  CIRCLE_PROJECT_REPONAME=homebrew-xorg
  CIRCLE_PROJECT_USERNAME=Linuxbrew
  CIRCLE_PR_NUMBER=350
  CIRCLE_PR_REPONAME=homebrew-xorg
  CIRCLE_PR_USERNAME=maxim-belkin
  CIRCLE_PULL_REQUEST=https://github.com/Linuxbrew/homebrew-xorg/pull/350
  CIRCLE_PULL_REQUESTS=https://github.com/Linuxbrew/homebrew-xorg/pull/350
  CIRCLE_REPOSITORY_URL=git@github.com:Linuxbrew/homebrew-xorg.git
  CIRCLE_SHA1=5c20503efe490225552bff3cde2837efa606c3d8
  CIRCLE_SHELL_ENV=/tmp/.bash_env-5a99a195c9e77c000191763d-0-build
  CIRCLE_STAGE=build
  CIRCLE_USERNAME=maxim-belkin
  CIRCLE_WORKING_DIRECTORY=~/project
  CI_PULL_REQUEST=https://github.com/Linuxbrew/homebrew-xorg/pull/350
  CI_PULL_REQUESTS=https://github.com/Linuxbrew/homebrew-xorg/pull/350
```

.circleci/config.yml
sjackman: I'm okay with `Linuxbrew`.

maxim-belkin: and what about `CIRCLE_PROJECT_REPONAME`? 
```
$ echo -n '$CIRCLE_PROJECT_REPONAME' | wc -c
24
$ echo -n 'homebrew-xorg' | wc -c
13
```

.circleci/config.yml
```diff
@@ -4,31 +4,36 @@ jobs:
...
+            cd $CIRCLE_PROJECT_REPONAME
+            if [[ -n "$CIRCLE_PR_NUMBER" ]]; then
+                git fetch origin pull/$CIRCLE_PR_NUMBER/head
+            fi
+            git reset --hard $CIRCLE_SHA1
+      - run:
```
maxim-belkin, line 27: How about:
```bash
            if [[ -n "$CIRCLE_PR_NUMBER" ]]; then
                git pull origin pull/$CIRCLE_PR_NUMBER/head
            fi
```
?

.circleci/config.yml
maxim-belkin: Hmmm... I guess we can keep `CIRCLE_PROJECT_REPONAME` because this way file can be used for all Linuxbrew repos

.circleci/config.yml
maxim-belkin: We can also use
```
$CIRCLE_BRANCH/head
``` 
instead of
```
pull/$CIRCLE_PR_NUMBER/head
```

sjackman: `pull/$CIRCLE_PR_NUMBER/head` is more clear to me. `$CIRCLE_BRANCH/head` doesn't work when `$CIRCLE_BRANCH` is `master`. Neither does `pull/$CIRCLE_PR_NUMBER/head`, but it's more clear why it doesn't work.

sjackman: `git pull` creates a merge commit, which we definitely don't want.

maxim-belkin: default is `--ff-only`, so it would be fine

maxim-belkin: no, default is `-ff`, but in this case it should work anyway :)

sjackman: One option may be…
```
git fetch origin $CIRCLE_SHA1:$CIRCLE_BRANCH
git checkout $CIRCLE_BRANCH
```
I think I still prefer `git reset --hard $CIRCLE_SHA1` over `git checkout $CIRCLE_BRANCH`

sjackman: The default behaviour of `git pull` is to create a merge commit, which we don't want.

maxim-belkin:
```
$ git help pull
...
       --ff
           When the merge resolves as a fast-forward, only update the branch pointer, without creating a merge commit. This is the default behavior.
```

sjackman: `When the merge resolves as a fast-forward`, when it doesn't, it creates a merge commit, which we don't want.

maxim-belkin: I'm fine with resetting as it brings us to the same place...
but `git reset --hard` usually signifies that something has gone wrong and we're trying to fix it.

I thought PRs always resolve in fast-forwards... don't they? :confused: If they don' - sure, let's keep `reset --hard`. But if they are supposed to resolve in fast-forwards,then  we can do `--ff-only` which will fail if a merge commit is necessary and, therefore, it will serve as an additional verification step.

sjackman: If the PR is say one commit behind `origin/master`, then it won't resolve in a fast-forward. We can ask the submitter to rebase onto master, but it's nice not to have to if it's not needed.

maxim-belkin: Ah! Gotcha! OK, I'm leaving this part untouched.

maxim-belkin:
Thank you, Shaun (@sjackman), for a thorough review!

sjackman:
No worries, Maxim. I'll probably remove CircleCI `checkout` from other taps when I get to it.

PR #355, opened 2018-03-07
CircleCI: Disable analytics

maxim-belkin:
I don't think we want Homebrew analytics tracking our failing builds on CircleCI :trollface:


sjackman:
I believe analytics are disabled while `brew test-bot` is running, but yes, good idea to disable it here if you're running `brew update` outside of `brew test-bot`.

maxim-belkin:
Yes, I noticed analytics here: https://circleci.com/gh/Linuxbrew/homebrew-xorg/996 in `brew update` state

PR #361, opened 2018-03-07
More changes related to dependencies being actually build-time dependencies

maxim-belkin:
Some dependencies are, actually, build-time dependencies. Fixing this here.

### DIFF
--- a/Formula/libvdpau.rb
+++ b/Formula/libvdpau.rb
@@ -17,7 +17,7 @@ class Libvdpau < Formula
   # Required
   depends_on "linuxbrew/xorg/dri2proto" => :build
   depends_on "linuxbrew/xorg/libxext"
-  depends_on "linuxbrew/xorg/libx11"
+  depends_on "linuxbrew/xorg/libx11" => :build
 
   # optional
   depends_on "doxygen" => :optional
--- a/Formula/libxdamage.rb
+++ b/Formula/libxdamage.rb
@@ -15,7 +15,7 @@ class Libxdamage < Formula
 
   depends_on "pkg-config" => :build
   depends_on "linuxbrew/xorg/damageproto" => :build
-  depends_on "linuxbrew/xorg/libxfixes"
+  depends_on "linuxbrew/xorg/libxfixes" => :build
   depends_on "linuxbrew/xorg/fixesproto" => :build
   depends_on "linuxbrew/xorg/xextproto" => :build
   depends_on "linuxbrew/xorg/libx11"
--- a/Formula/libxi.rb
+++ b/Formula/libxi.rb
@@ -17,7 +17,7 @@ class Libxi < Formula
 
   depends_on "pkg-config" => :build
 
-  depends_on "linuxbrew/xorg/libxfixes"
+  depends_on "linuxbrew/xorg/libxfixes" => :build
   depends_on "linuxbrew/xorg/xextproto" => :build
   depends_on "linuxbrew/xorg/libxext"
   depends_on "linuxbrew/xorg/libx11"
--- a/Formula/libxtst.rb
+++ b/Formula/libxtst.rb
@@ -17,7 +17,7 @@ class Libxtst < Formula
 
   depends_on "linuxbrew/xorg/util-macros" => :build
   depends_on "pkg-config" => :build
-  depends_on "linuxbrew/xorg/libxi"
+  depends_on "linuxbrew/xorg/libxi" => :build
 
   # Patch for xmlto
   patch do
--- a/Formula/libxcb.rb
+++ b/Formula/libxcb.rb
@@ -17,7 +17,7 @@ class Libxcb < Formula
   depends_on "pkg-config" => :build
   depends_on "linuxbrew/xorg/libxau"
   depends_on "linuxbrew/xorg/xcb-proto" => :build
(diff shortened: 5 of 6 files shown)


maxim-belkin:
Requesting your review, @sjackman. Basically, I just want an additional pair of eyes to check that `brew linkage <formulae-in-this-pr>` report formulae with no linkage that I modify to `:build` here.

Formula/libvdpau.rb
```diff
@@ -17,7 +17,7 @@ class Libvdpau < Formula
   # Required
   depends_on "linuxbrew/xorg/dri2proto" => :build
   depends_on "linuxbrew/xorg/libxext"
-  depends_on "linuxbrew/xorg/libx11"
+  depends_on "linuxbrew/xorg/libx11" => :build
```
sjackman, line 20: Is it needed at all? Is it used only for its headers, or is it statically linked?

sjackman: The same question goes for each of these.

maxim-belkin: built libvdpau from source: https://gist.github.com/86c47f853abf03cc35d78653ead6ec97

sjackman: So, `libvdpau` does not depend on `libx11` then?

maxim-belkin: 1. In `/home/linuxbrew/.linuxbrew/Cellar/libvdpau/1.1.1/include/vdpau/vdpau_x11.h` there is  `#include <X11/Xlib.h>`
2. In `/home/linuxbrew/.linuxbrew/Cellar/libvdpau/1.1.1/lib/pkgconfig/vdpau.pc` there is `Requires.private: x11`

So, it needs `libx11`, it's just not linked to it... Does this secure a standard dependency type?

Formula/libxdamage.rb
```diff
@@ -15,7 +15,7 @@ class Libxdamage < Formula
 
   depends_on "pkg-config" => :build
   depends_on "linuxbrew/xorg/damageproto" => :build
-  depends_on "linuxbrew/xorg/libxfixes"
+  depends_on "linuxbrew/xorg/libxfixes" => :build
```
maxim-belkin, line 18:
```
~/.linuxbrew/Cellar/libxdamage/1.1.4/lib/pkgconfig/xdamage.pc:Requires: xproto damageproto >= 1.1 xfixes
~/.linuxbrew/Cellar/libxdamage/1.1.4/lib/pkgconfig/xdamage.pc:Requires.private: x11 xfixes
~/.linuxbrew/Cellar/libxdamage/1.1.4/include/X11/extensions/Xdamage.h:#include <X11/extensions/Xfixes.h>
```

Formula/libxi.rb
```diff
@@ -17,7 +17,7 @@ class Libxi < Formula
 
   depends_on "pkg-config" => :build
 
-  depends_on "linuxbrew/xorg/libxfixes"
+  depends_on "linuxbrew/xorg/libxfixes" => :build
```
maxim-belkin, line 20:
```
~/.linuxbrew/Cellar/libxi/1.7.9/lib/pkgconfig/xi.pc:Requires.private: x11 xext xfixes
~/.linuxbrew/Cellar/libxi/1.7.9/include/X11/extensions/XInput2.h:#include <X11/extensions/Xfixes.h> /* PointerBarrier */
```

Formula/libxtst.rb
```diff
@@ -17,7 +17,7 @@ class Libxtst < Formula
 
   depends_on "linuxbrew/xorg/util-macros" => :build
   depends_on "pkg-config" => :build
-  depends_on "linuxbrew/xorg/libxi"
+  depends_on "linuxbrew/xorg/libxi" => :build
```
maxim-belkin, line 20:
```
~/.linuxbrew/Cellar/libxtst/1.2.3/lib/pkgconfig/xtst.pc:Requires.private: x11 xext xextproto xi
~/.linuxbrew/Cellar/libxtst/1.2.3/include/X11/extensions/XTest.h:#include <X11/extensions/XInput.h>
```

maxim-belkin:
I just got an idea! There has to be:
```
depends_on "other_formula" => :no_linkage
```
!!!

Formula/libvdpau.rb
sjackman: That's pretty common for some libraries, like `boost`. It's standard `:build` dependency.

sjackman:
That's the same as `depends_on "other_formula" => :build`

maxim-belkin:
> That's the same as `depends_on "other_formula" => :build`

Of course not.

`:build` dependencies are needed only when the formula is built (`automake`)
they are not needed at runtime.

[not-yet-existing] `:no_linkage` dependencies are needed when the formula is built and at runtime, but are not linked to libraries of the formula (_e.g._, formula uses header files provided  by the dependency).

maxim-belkin:
This is the situation that we have here. `configure` script checks that dependency is installed and fails if it is not. But libraries that are built do not link to the libraries of the dependency: only header files are used.

sjackman:
If a formula does not link to the libraries of its dependency, and does not execute the executables of its dependency at run time, then the dependency is `:build`.

maxim-belkin:
Closing this PR because according to upstream, these dependencies should be standard (normal) dependencies, at least at this moment.

PR #369, opened 2018-03-11
libdrm: fix deps

maxim-belkin:
Fix couple of dependencies in `libdrm`.


maxim-belkin:
```
==> brew linkage --test linuxbrew/xorg/libdrm
System libraries:
  /lib/x86_64-linux-gnu/libc.so.6
  /lib/x86_64-linux-gnu/libm.so.6
No broken dylib links
No unwanted system libraries
```

PR #370, opened 2018-03-16
Fixing the url for xcb-proto patches.

marlukcz:
Fixing the url for xcb-proto patches.

### DIFF
--- a/Formula/xcb-proto.rb
+++ b/Formula/xcb-proto.rb
@@ -19,12 +19,12 @@ class XcbProto < Formula
   depends_on "python@2" => :build unless which "python2.7"
 
   patch :p1 do
-    url "http://www.linuxfromscratch.org/patches/blfs/svn/xcb-proto-1.12-python3-1.patch"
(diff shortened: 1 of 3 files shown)


marlukcz:
Any idea why the build failed ?

maxim-belkin:
Travis failed to install `gcc-5`.
Would you mind copying the patch to `/Patches` directory instead of just pointing to the new location?
Also, please add a comment right before the patch url where to find it
```ruby
patch :p1 do
  # March 16, 2018
  # http://www.linuxfromscratch.org/patches/blfs/8.2/xcb-proto-1.12-python3-1.patch
  url "https://github.com/Linuxbrew/homebrew-xorg/master/Patches/xcb-proto-1.12-python3-1.patch"
end
```

marlukcz:
Yep, it won't build successfully until merged. Should I create a separate pull request for adding files and for changing URL's ? Original URL's are invalid as well ...

maxim-belkin:
Nope, no need to. It's good enough! thank you

maxim-belkin:
Thank you @marlukcz!

PR #379, opened 2018-07-24
mesa: Add option of "without-gpu"

Gwang-Jin:
For the system without graphic driver, I would like to add an option "with-only-cpu".

- [x] Have you included a log of the failed build without your patch using `brew gist-logs <formula>`?

### DIFF
--- a/Formula/mesa.rb
+++ b/Formula/mesa.rb
@@ -114,13 +114,13 @@ class Mesa < Formula
 
     if build.with? "gpu"
       args += %W[
-        --with-egl-platforms=drm,x11,surfaceless#{build.with?("wayland") ? ",wayland" : ""}
+        --with-platforms=drm,x11,surfaceless#{build.with?("wayland") ? ",wayland" : ""}
         --with-gallium-drivers=i915,nouveau,r300,r600,radeonsi,svga,swrast,swr
         --with-dri-drivers=i965,nouveau,radeon,r200,swrast
       ]
     else
       args += %W[
-        --with-egl-platforms=
+        --with-platforms=
         --with-gallium-drivers=swrast,swr
         --with-dri-drivers=
       ]
--- a/Formula/mesa.rb
+++ b/Formula/mesa.rb
@@ -3,7 +3,7 @@ class Mesa < Formula
   homepage "https://dri.freedesktop.org"
   url "https://mesa.freedesktop.org/archive/mesa-17.2.3.tar.xz"
   sha256 "a0b0ec8f7b24dd044d7ab30a8c7e6d3767521e245f88d4ed5dd93315dc56f837"
-  revision 2
+  revision 3
 
   bottle do
     root_url "https://linuxbrew.bintray.com/bottles-xorg"
--- a/Formula/mesa.rb
+++ b/Formula/mesa.rb
@@ -13,6 +13,7 @@ class Mesa < Formula
 
   option "without-test", "Skip compile-time tests"
   option "with-static", "Build static libraries (not recommended)"
+  option "without-gpu", "Build without graphics hardware"
 
   depends_on "pkg-config" => :build
   depends_on "python" => :build
@@ -77,6 +78,9 @@ class Mesa < Formula
       system "python", *Language::Python.setup_install_args(libexec/"vendor")
     end
 
+    gpu = build.with?("gpu") ? "yes" : "no"
+    nogpu = build.with?("gpu") ? "no" : "yes"
+
     args = %W[
       CFLAGS=#{ENV.cflags}
       CXXFLAGS=#{ENV.cflags}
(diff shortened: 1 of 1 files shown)


Formula/mesa.rb
```diff
@@ -110,6 +111,38 @@ def install
...
+      --with-dri-drivers=
+      --disable-egl
+      --with-egl-platforms=
+      --disable-gbm
+      --disable-glx
+      --disable-osmesa
+      --enable-gallium-osmesa
+      --with-llvm-prefix=#{Formula["llvm@4"].opt_prefix}
+      ]
+    end
```
maxim-belkin: Instead of redefining `args` completely, determine common arguments and leave them in the original definition and then add
```
if `build.with? "gpu"`
...
else
...
end
```

Formula/mesa.rb
```diff
@@ -13,6 +13,7 @@ class Mesa < Formula
 
   option "without-test", "Skip compile-time tests"
   option "with-static", "Build static libraries (not recommended)"
+  option "with-only-cpu", "without graphics hardware, and OSMesa"
```
maxim-belkin: rename this to `"without-gpu"`.
Capitalize  the first word in the explanation: `Build without GPU hardware`.

Does it worth splitting it into two options? `without-gpu` and `without-osmesa`?

Formula/mesa.rb
```diff
@@ -121,10 +154,10 @@ def install
 
     system "./autogen.sh", *args
     system "make"
-    system "make", "-C", "xdemos", "DEMOS_PREFIX=#{prefix}"
+    system "make", "-C", "xdemos", "DEMOS_PREFIX=#{prefix}" if build.without?("only-cpu")
```
maxim-belkin: after changing the option to `without-gpu`, change the conditional to `if build.with? "gpu"`. No parentheses.

Formula/mesa.rb
```diff
@@ -121,10 +154,10 @@ def install
...
     system "make", "check" if build.with?("test")
     system "make", "install"
-    system "make", "-C", "xdemos", "DEMOS_PREFIX=#{prefix}", "install"
+    system "make", "-C", "xdemos", "DEMOS_PREFIX=#{prefix}", "install" if build.without?("only-cpu")
```
maxim-belkin: Same as above

maxim-belkin (changes requested):
Thanks for the PR, @Gwang-Jin! Please see my comments

Formula/mesa.rb
Gwang-Jin: ``without-gnu`` is OK. But ``without-osmesa`` is not my intention. Maybe ``--enable-gallium-osmesa`` would be better in favor of the newer "gallium" OSMesa.

Gwang-Jin:
I succeeded in my system
`brew install mesa` and `brew install mesa --without-gpu`
But, I don't know why `continuous-integration/travis-ci/pr` gives an error message:
````
==> brew install gcc
==> FAILED
systemd cannot be built with any available compilers.
````

Can't `systemd` be built in current Linuxbrew? Any suggestions would be appreciated.

File: Formula/mesa.rb (hunk 1 above)
Comment (Gwang-Jin) [reply Formula/mesa.rb]: Ok. I did.

**Comment by maxim-belkin** [comment]
Don't worry about Travis CI.

File: Formula/mesa.rb (hunk 3 above)
Comment (Gwang-Jin) [reply Formula/mesa.rb]: Ok.

File: Formula/mesa.rb (hunk 4 above)
Comment (Gwang-Jin) [reply Formula/mesa.rb]: Ok

File: Formula/mesa.rb (hunk 2 above)
Comment (Gwang-Jin) [reply Formula/mesa.rb]: I changed the explanation to "Build without graphics hardware".

**Comment by Gwang-Jin** [comment]
@maxim-belkin I think that your all comments was reflected.

**Comment by maxim-belkin** [comment]
According to `configure`, `with-egl-platforms` is deprecated. We need to address this

```
  --with-egl-platforms[=DIRS...]
                          DEPRECATED: use --with-platforms
                          instead[default=auto]
  --with-platforms[=DIRS...]
                          comma delimited native platforms libEGL/Vulkan/other
                          supports, e.g. "x11,drm,wayland,surfaceless..."
                          [default=auto]
```

**Comment by maxim-belkin** [comment]
I took care of the deprecated `with-egl-platforms` option

**Comment by maxim-belkin** [comment]
```ruby
==> brew linkage --test linuxbrew/xorg/mesa
No broken library linkage
Warning: linuxbrew/xorg/mesa dependency gmp was built with a different C++ standard
library (libstdc++ from gcc-5). This may cause problems at runtime.
```
@sjackman shall we add a dependency on `gcc` to `mesa`? 

For reference:

```
$ brew uses mesa
fltk
glew
libepoxy
linuxbrew/xorg/glu
linuxbrew/xorg/xdriinfo
mednafen
mupdf
qt
vtk
```

Formula/mesa.rb
```diff
@@ -3,16 +3,11 @@ class Mesa < Formula
   homepage "https://dri.freedesktop.org"
   url "https://mesa.freedesktop.org/archive/mesa-17.2.3.tar.xz"
   sha256 "a0b0ec8f7b24dd044d7ab30a8c7e6d3767521e245f88d4ed5dd93315dc56f837"
-  revision 2
-
-  bottle do
-    root_url "https://linuxbrew.bintray.com/bottles-xorg"
-    rebuild 1
-    sha256 "b4c2ae6373c90eb23f5e8ed33d9757268802b793d56b17d6d58dd44ad9a48e32" => :x86_64_linux
-  end
```
sjackman:
```
linuxbrew/xorg/mesa:
  * `bottle` is not defined
```
https://circleci.com/gh/Linuxbrew/homebrew-xorg/1091

Don't remove the bottle block.

sjackman:
> shall we add a dependency on gcc to mesa?

No, probably not. What problem are you trying to address?

maxim-belkin:
> > shall we add a dependency on gcc to mesa?
>
> No, probably not. What problem are you trying to address?

```
Warning: linuxbrew/xorg/mesa dependency gmp was built with a different C++ standard
library (libstdc++ from gcc-5). This may cause problems at runtime.
```

Regarding the bottle: I _think_ we need to bump up the revision because of this PR (too many small changes). So, I can just bump up the `revision` and keep the bottle block?

sjackman:
You can generally ignore that warning.

When you bump the revision, you can just leave the bottle block as is.

maxim-belkin:
Ok-dok! thanks

maxim-belkin:
Merged!
Thank you, @Gwang-Jin, for contributing to homebrew-xorg!

Gwang-Jin:
@maxim-belkin Wow, thank you for helping me.

PR #351, opened 2018-03-05
mesa: attempt to avoid dependency on llvm

maxim-belkin:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Linuxbrew/homebrew-xorg/blob/master/.github/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Linuxbrew/homebrew-xorg/pulls) for the same update/change?
- [x] Does your submission pass
`brew audit --strict <formula>` (where `<formula>` is the name of the formula you're submitting)?
- [x] Have you built your formula locally prior to submission with `brew install <formula>`?

Related to #329

PR #383, opened 2018-08-17
mesa: mostly stylistic changes

maxim-belkin:
Mostly stylistic changes to mesa formula.

@sjackman, could you please review lines 150 and 151 of this PR?


maxim-belkin:
These changes have been ported from #351

sjackman:
```diff
-           --sysconfdir=#{etc}
-           --localstatedir=#{var}
+           --sysconfdir=#{Formula["libva"].opt_prefix}/etc
+           --localstatedir=#{Formula["libva"].opt_prefix}/var
```
If those files are intended to be modified by the user, it should be the former. If they should not be modified, then the latter.

PR #396, opened 2018-09-07
mesa: fix up dependencies

maxim-belkin:
A follow-up to Linuxbrew/homebrew-xorg#351

### DIFF
--- a/Formula/mesa.rb
+++ b/Formula/mesa.rb
@@ -31,18 +31,23 @@ class Mesa < Formula
 
   depends_on "linuxbrew/xorg/libdrm"
   depends_on "systemd" # provides libudev <= needed by "gbm"
-  depends_on "llvm@4" # failed with llvm@6
+  depends_on "llvm@4" => :build # failed with llvm@6
   depends_on "libelf" # radeonsi requires libelf when using llvm
   depends_on "linuxbrew/xorg/libomxil-bellagio"
-  depends_on "linuxbrew/xorg/wayland-protocols" => :recommended
+  depends_on "linuxbrew/xorg/wayland-protocols" => [:recommended, :build]
+  depends_on "linuxbrew/xorg/xproto" => :build
   depends_on "valgrind" => :recommended
   depends_on "linuxbrew/xorg/libglvnd" => :optional
   depends_on "linuxbrew/xorg/libva" => :recommended
-  depends_on "linuxbrew/xorg/libvdpau"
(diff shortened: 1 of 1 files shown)


Formula/mesa.rb
```diff
@@ -31,18 +31,23 @@ class Mesa < Formula
...
   depends_on "libelf" # radeonsi requires libelf when using llvm
   depends_on "linuxbrew/xorg/libomxil-bellagio"
-  depends_on "linuxbrew/xorg/wayland-protocols" => :recommended
+  depends_on "linuxbrew/xorg/wayland-protocols" => [:recommended, :build]
+  depends_on "linuxbrew/xorg/xproto" => :build
   depends_on "valgrind" => :recommended
   depends_on "linuxbrew/xorg/libglvnd" => :optional
   depends_on "linuxbrew/xorg/libva" => :recommended
-  depends_on "linuxbrew/xorg/libvdpau"
+  depends_on "linuxbrew/xorg/libvdpau" => :build
```
sjackman, line 42: Why no run-time linkage, is it header only? Is this dependency required?

maxim-belkin: `libvdpau` is not header-only:
```
/home/linuxbrew/.linuxbrew/Cellar/libvdpau/1.1.1/lib/libvdpau.so
/home/linuxbrew/.linuxbrew/Cellar/libvdpau/1.1.1/lib/vdpau/libvdpau_trace.so
```
But mesa does not link against these. I _think_ `mesa` uses `libvdpau` to build libs like

```
/home/linuxbrew/.linuxbrew/Cellar/mesa/17.2.3_3/lib/vdpau/libvdpau_r300.so
/home/linuxbrew/.linuxbrew/Cellar/mesa/17.2.3_3/lib/vdpau/libvdpau_radeonsi.so
/home/linuxbrew/.linuxbrew/Cellar/mesa/17.2.3_3/lib/vdpau/libvdpau_nouveau.so
```

maxim-belkin: > Is this dependency required?

yes, it is required - the build fails without it.

maxim-belkin: I see several 
```
-I/home/linuxbrew/.linuxbrew/Cellar/libvdpau/1.1.1/include
```
in `mesa` build logs... perhaps, mesa uses header files only, actually...

sjackman: 👍 That's possible.

sjackman:
Thanks, Maxim!

PR #448, opened 2018-10-04
xorgproto: new formula

maxim-belkin:
New formula that will replace a bunch of other `proto` packages.

### DIFF
--- a/Formula/xorgproto.rb
+++ b/Formula/xorgproto.rb
@@ -1,0 +1,38 @@
+class Xorgproto < Formula
+  desc "X.Org Protocol Headers"
+  homepage "https://www.x.org/"
+  url "https://xorg.freedesktop.org/archive/individual/proto/xorgproto-2018.4.tar.bz2"
+  sha256 "fee885e0512899ea5280c593fdb2735beb1693ad170c22ebcc844470eec415a0"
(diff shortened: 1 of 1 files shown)


sjackman:
Woohoo! That's great! We'll have to test the transition for the old many-packages-system to the new system so that all the old `*proto` packages get unlinked and replaced by `xorgproto` by `brew update && brew upgrade`.

sjackman:
I don't know myself how to handle those migrations. Could you please research it?

maxim-belkin:
Hmm... Those are just `proto` packages, so nothing should change except for dependency definitions in formulae. Or am I missing something?

sjackman:
I'm guessing that `xorgproto` installs files that have the same name and conflict with the old `*proto` formulae. When the user runs `brew upgrade`, it'll install `xorgproto`, and attempt to link it, and fail due to the conflicts. `brew update && brew upgrade` needs to unlink all the old `*proto` packages for the user.

maxim-belkin:
ah, that's right... Time to 🤔

sjackman:
I believe there's some kind of support in Homebrew/brew for this sort of case, though I've never used it. It might have the name "migrate" something or other.

PR #455, opened 2019-02-14
xcb: Simplify dependency tree

iMichka:
These depencies are only needed for the tests, and make the dependency
tree too big (and create circular dependencies for tcl-tk and python).

Remove the test too, as we almost never run tests a build time.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,11 +11,6 @@ jobs:
       HOMEBREW_VERBOSE_USING_DOTS: 1
     steps:
       - run:
-          name: Install minimal python for libxcb build
-          command: |
-            sudo apt-get update
-            sudo apt-get install -y --no-install-recommends python2.7-minimal
-      - run:
           name: Set up Git
           command: |
             git config --global user.name LinuxbrewTestBot
--- a/Formula/libxcb.rb
+++ b/Formula/libxcb.rb
@@ -3,12 +3,12 @@ class Libxcb < Formula
   homepage "https://www.x.org/" ### http://www.linuxfromscratch.org/blfs/view/svn/x/x7lib.html
   url "https://xcb.freedesktop.org/dist/libxcb-1.13.1.tar.bz2"
   sha256 "a89fb7af7a11f43d2ce84a844a4b38df688c092bf4b67683aef179cdf2a647c4"
+  revision 1
   # tag "linuxbrew"
 
   bottle do
     root_url "https://linuxbrew.bintray.com/bottles-xorg"
     cellar :any_skip_relocation
-    sha256 "657c38f163f14bc65729f6b15414da87d896adc3d005642289bd9d4c0544b05f" => :x86_64_linux
   end
 
   option "without-test", "Skip compile-time tests"
@@ -17,7 +17,6 @@ class Libxcb < Formula
 
   depends_on "linuxbrew/xorg/xcb-proto" => :build
   depends_on "pkg-config" => :build
-  depends_on "python" => :build
   depends_on "linuxbrew/xorg/libpthread-stubs" # xcb.pc references pthread-stubs
   depends_on "linuxbrew/xorg/libxau"
   depends_on "linuxbrew/xorg/libxdmcp"
@@ -25,11 +24,6 @@ class Libxcb < Formula
   if build.with? "devel-docs"
     depends_on "doxygen" => :build
     depends_on "graphviz" => :build
-  end
-
-  if build.with? "test"
-    depends_on "check" => :build
-    depends_on "libxslt" => [:build, :recommended]
   end
 
   def install
@@ -53,7 +47,6 @@ class Libxcb < Formula
 
     system "./configure", *args
     system "make"
-    system "make", "check" if build.with? "test"
     system "make", "install"
   end
 end
--- a/Formula/xcb-proto.rb
+++ b/Formula/xcb-proto.rb
@@ -3,21 +3,18 @@ class XcbProto < Formula
   homepage "https://www.x.org/"
   url "https://xcb.freedesktop.org/dist/xcb-proto-1.13.tar.bz2"
   sha256 "7b98721e669be80284e9bbfeab02d2d0d54cd11172b72271e47a2fe875e2bde1"
+  revision 1
   # tag "linuxbrew"
 
   bottle do
     root_url "https://linuxbrew.bintray.com/bottles-xorg"
     cellar :any_skip_relocation
-    sha256 "68f456b5eeeb486d33019d410ab25acc3cdc8fb10240c4c08f3b59d4e5bf5d1e" => :x86_64_linux
   end
 
-  option "without-test", "Skip compile-time tests"
-  option "with-python@2", "Build with Python 2"
-
-  depends_on "libxml2" => :build if build.with? "test"
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
   depends_on "pkg-config" => :build
-  depends_on "python" => :build if build.without? "python@2"
-  depends_on "python@2" => [:build, :optional]
+  depends_on "minimal-python"
 
   def install
     args = %W[
@@ -27,9 +24,9 @@ class XcbProto < Formula
       --disable-silent-rules
     ]
 
+    system "./autogen.sh"
     system "./configure", *args
     system "make"
-    system "make", "check" if build.with? "test"
     system "make", "install"
   end
 end


maxim-belkin:
no objections if the build succeeds

maxim-belkin (approved):

iMichka:
> The command "gem install bundler -v '< 2'" failed and exited with 1 during .

The travis error is unrelated.

iMichka:
Bottle upload fails with:
```
! [remote rejected] master -> master (shallow update not allowed)
error: failed to push some refs to 'git@github.com:LinuxbrewTestBot/homebrew-xorg.git'
Error: Failure while executing; `git push --force git@github.com:LinuxbrewTestBot/homebrew-xorg.git master:master :refs/tags/pr-455` exited with 1.
```

maxim-belkin:
There is no pr-455 branch/release/whatever in LinuxbrewTestBot.
https://github.com/LinuxbrewTestBot/homebrew-xorg/compare/Linuxbrew:master...pr-455.patch

iMichka:
I tried to upload the bottle manually, exactly like I do for homebrew-core, and it failed too:
```
CIRCLE_PR_NUMBER=455 /home/linuxbrew/.linuxbrew/bin/brew test-bot --ci-upload --bintray-org=linuxbrew --git-name=LinuxbrewTestBot --git-email=testbot@linuxbrew.sh
Homebrew/homebrew-test-bot 
ARGV: --ci-upload --bintray-org=linuxbrew --git-name=LinuxbrewTestBot --git-email=testbot@linuxbrew.sh
git checkout -f master
Already on 'master'
Your branch is up to date with 'origin/master'.
git reset --hard origin/master
HEAD is now at 2a6a107 Switch to using xorgproto
brew update
Already up-to-date.
brew pull --clean https://github.com/Linuxbrew/homebrew-xorg/pull/455
==> Fetching patch 
Patch: https://github.com/Linuxbrew/homebrew-xorg/pull/455.patch
#=#=-#  #                                                                     
==> Applying patch
Applying: xcb-proto: Do not depend on Python and libxml2
==> Patch changed:
 Formula/xcb-proto.rb | 9 +--------
 1 file changed, 1 insertion(+), 8 deletions(-)
brew bottle --merge --write xcb-proto--1.13_1.x86_64_linux.bottle.json
==> linuxbrew/xorg/xcb-proto
  bottle do
    root_url "https://linuxbrew.bintray.com/bottles-xorg"
    cellar :any_skip_relocation
    sha256 "96594728b55ff238ffd673817185caa4a60b1de18c6f33642cb0411026700035" => :x86_64_linux
  end
[master fef8732] xcb-proto: update 1.13_1 bottle.
 1 file changed, 1 insertion(+)
git push --force git@github.com:LinuxbrewTestBot/homebrew-xorg.git master:master :refs/tags/pr-455
Enumerating objects: 4749, done.
Counting objects: 100% (4749/4749), done.
Delta compression using up to 8 threads
Compressing objects: 100% (930/930), done.
Writing objects: 100% (4749/4749), 836.86 KiB | 8.20 MiB/s, done.
Total 4749 (delta 3821), reused 4731 (delta 3806)
remote: Resolving deltas: 100% (3821/3821), done.
remote: warning: Deleting a non-existent ref.
To github.com:LinuxbrewTestBot/homebrew-xorg.git
 ! [remote rejected] master -> master (permission denied)
 ! [remote rejected] pr-455 (permission denied)
error: failed to push some refs to 'git@github.com:LinuxbrewTestBot/homebrew-xorg.git'
Error: Failure while executing; `git push --force git@github.com:LinuxbrewTestBot/homebrew-xorg.git master:master :refs/tags/pr-455` exited with 1.
```

iMichka:
I also need to double check, but it still tries to setup some python stuff:
> xcb-proto/1.13_1/lib/python3.5/site-packages/xcbgen

maxim-belkin:
Python 3.5 that is picked up above is provided by the linuxbrew/linuxbrew image (`python3-minimal` package). `xcb-proto` requires Python 2.5+ in order for install to succeed. So, the question is how badly do you need to get rid of it?

maxim-belkin (changes requested):
Retracting my approval for now.
`xcb-proto` requires Python 2.5+ so there is no way to build this formula without any Python
Note: `libxml2` that is used in `make check` depends on `python`.

Formula/xcb-proto.rb
```diff
@@ -3,21 +3,15 @@ class XcbProto < Formula
   homepage "https://www.x.org/"
   url "https://xcb.freedesktop.org/dist/xcb-proto-1.13.tar.bz2"
   sha256 "7b98721e669be80284e9bbfeab02d2d0d54cd11172b72271e47a2fe875e2bde1"
+  revision 1
   # tag "linuxbrew"
 
   bottle do
     root_url "https://linuxbrew.bintray.com/bottles-xorg"
     cellar :any_skip_relocation
-    sha256 "68f456b5eeeb486d33019d410ab25acc3cdc8fb10240c4c08f3b59d4e5bf5d1e" => :x86_64_linux
```
maxim-belkin, line 11: This deletion is not needed because `test-bot` is executed without `--keep-old`

iMichka:
```
  * C: 15: col 3: FormulaAudit/DependencyOrder: dependency "autoconf" (line 15) should be put before dependency "pkg-config" (line 14)
  * C: 16: col 3: FormulaAudit/DependencyOrder: dependency "automake" (line 16) should be put before dependency "pkg-config" (line 14)
```

I'll fix this when pulling the patch. The build lasted 7 hours, so I would like to avoid running it again.

> failed: brew fetch --retry plotutils

I need to rebuild that bottle, this is a separate thing

```
failed: brew install --only-dependencies --include-test mesa

        Error: Operation already in progress for mesa.formula
Another active Homebrew process is already using mesa.formula.
Please wait for it to finish or terminate it to continue.
```

This is an issue because we have 2 mesa formulae. The one in core depends on "freeglut" for the tests, which creates a dependency cycle between the two mesa formulae. My plan is to get rid of the `mesa` formula in xorg, and to fix the `mesa` formula in core.

iMichka:
Re-running this without the python-minimal package from apt. Just to make sure we are doing the right thing here.

maxim-belkin:
> My plan is to get rid of the mesa formula in xorg, and to fix the mesa formula in core.

I don't think that's a good idea. Our formula existed long before the one in core.

iMichka:
> I don't think that's a good idea. Our formula existed long before the one in core.

Let's discuss this later :) We could also drop the core formula. We will see.

Coming back to this PR: it is ready to be reviewed.

iMichka:
Some more details about this:

`xcb-proto` installs some plain python files, which are python-agnostic. Nothing is compiled here, and nothing links against xcb-proto. The files are installed in:
/home/linuxbrew/.linuxbrew/Cellar/xcb-proto/1.13/lib/python3.7/site-packages/xcbgen/

`pkg-config --variable=pythondir xcb-proto` will output that path.

This path is defined by whatever Python was used to build xcb-proto. I used `minimal-python` for that. But it could be any other Python. The idea is that you use `pkg-config` to get the right path.

`libxcb` does not build without these files.

So the only way to make this work is to have a Python version installed on the system. Some people may not have Python installed. And having the base Python installed by apt on the CI is also not wanted.

`minimal-python` fills that gap. It does bring in almost all the dependencies from Python. We could trim it down (pull requests are welcome), to make it as minimal as possible. But in the end, you will almost always install Python anyway. So the dependencies will be needed anyway. This is why there is no harm to install these right away.

sjackman:
Why can't the brewed `python` be used with `xcb-proto`? Is it a circular dependency?

iMichka:
Right now it is not. But as soon as I will enable tcl-tk for Python, it will create a circular dependency. Tcl-tk should not depend on Python. Which means that xorg should not depend on Python. Even not at build time.

sjackman:
Okay, so the purpose of `minimal-python` is to have a version of `python` that does not depend on `tcl-tk` and `xorg`. Which formulae depend on `python` and expect to have `tcl-tk` that are not currently working? It's only a difference of naming, but could we go the other way and add a `python-tcl-tk` formula?

maxim-belkin:
At the moment, both `python` and `minimal-python` are sort of "python-**no**-tcl-tk" because `xorg` dependency is added only when `build.with? "tcl-tk"` and there is no such option.

Also, I would strongly advice against using `depends_on "xorg"` because it produces a monstrous dependency list. What you can do is build with `depends_on "xorg"` once and get the list of packages that it links against with `brew linkage`. You can also grep`config.log` for all `"/home/linuxbrew/.llinuxbrew"` paths (`grep -ho '/home/linuxbrew/.linuxbrew/.*' * | sort -u`) and see if any of the formulae are used but not linked against.

iMichka:
> Which formulae depend on python and expect to have tcl-tk that are not currently working?

None. But I got a few request during the last year from people surprised that there was no tcl-tk by default. I use `matplotlib` sometimes and the first thing it will do is to look for tcl-tk.
The mac version has tk support by default (see the Python test block on mac). We try to keep mac and linux behaviour aligned when possible.

> It's only a difference of naming, but could we go the other way and add a python-tcl-tk formula?

People (me included) expect to `brew install python` and get tcl-tk support. Also, `python-tcl-tk` would be keg-only, which makes it less nice to use.

> At the moment, both python and minimal-python are sort of "python-no-tcl-tk" because xorg dependency is added only when build.with? "tcl-tk" and there is no such option.

Because of the dependency cycle. As soon as this patch is merged, we should be able to enable tcl-tk by default.

> Also, I would strongly advice against using depends_on "xorg" because it produces a monstrous dependency list. .... 

Agreed on the strategy here. I want to trim down the list to the bare necessary.

.circleci/config.yml
```diff
@@ -10,11 +10,6 @@ jobs:
       HOMEBREW_VERBOSE: 1
       HOMEBREW_VERBOSE_USING_DOTS: 1
     steps:
-      - run:
-          name: Install minimal python for libxcb build
-          command: |
-            sudo apt-get update
-            sudo apt-get install -y --no-install-recommends python2.7-minimal
```
maxim-belkin, line 17: 👍

Formula/libxcb.rb
```diff
@@ -3,12 +3,12 @@ class Libxcb < Formula
   homepage "https://www.x.org/" ### http://www.linuxfromscratch.org/blfs/view/svn/x/x7lib.html
   url "https://xcb.freedesktop.org/dist/libxcb-1.13.1.tar.bz2"
   sha256 "a89fb7af7a11f43d2ce84a844a4b38df688c092bf4b67683aef179cdf2a647c4"
+  revision 1
```
maxim-belkin, line 6: 👍

Formula/libxcb.rb
```diff
@@ -3,12 +3,12 @@ class Libxcb < Formula
...
   # tag "linuxbrew"
 
   bottle do
     root_url "https://linuxbrew.bintray.com/bottles-xorg"
     cellar :any_skip_relocation
-    sha256 "657c38f163f14bc65729f6b15414da87d896adc3d005642289bd9d4c0544b05f" => :x86_64_linux
```
maxim-belkin, line 11: It's not necessary to delete this line in this repo (we don't use `--keep-old` flag here).

Formula/libxcb.rb
```diff
@@ -17,7 +17,6 @@ class Libxcb < Formula
 
   depends_on "linuxbrew/xorg/xcb-proto" => :build
   depends_on "pkg-config" => :build
-  depends_on "python" => :build
```
maxim-belkin, line 20: 👍

Formula/libxcb.rb
```diff
@@ -27,11 +26,6 @@ class Libxcb < Formula
     depends_on "graphviz" => :build
   end
 
-  if build.with? "test"
-    depends_on "check" => :build
-    depends_on "libxslt" => [:build, :recommended]
-  end
-
```
maxim-belkin, line 34: 👎 I'd prefer to keep the option of running `--with-test`.
You can switch `without-test` to `with-test` IMO if tests bother you. I'm also OK with removing/changing `:recommended` option for `libxslt`.

Formula/libxcb.rb
```diff
@@ -53,7 +47,6 @@ def install
 
     system "./configure", *args
     system "make"
-    system "make", "check" if build.with? "test"
```
maxim-belkin, line 56: related to my comment above

Formula/xcb-proto.rb
```diff
@@ -3,21 +3,18 @@ class XcbProto < Formula
   homepage "https://www.x.org/"
   url "https://xcb.freedesktop.org/dist/xcb-proto-1.13.tar.bz2"
   sha256 "7b98721e669be80284e9bbfeab02d2d0d54cd11172b72271e47a2fe875e2bde1"
+  revision 1
```
maxim-belkin, line 6: 👍

Formula/xcb-proto.rb
```diff
@@ -27,9 +24,9 @@ def install
       --disable-silent-rules
     ]
 
+    system "./autogen.sh"
```
maxim-belkin, line 27: I guess you checked that it does not compile without it, right?

maxim-belkin (commented):
And here are my comments!

Formula/xcb-proto.rb
sjackman: If this line is necessary, could you add a comment describing why, preferably with the error message that it fixes?

sjackman:
>> Which formulae depend on python and expect to have tcl-tk that are not currently working?

> None. But I got a few request during the last year from people surprised that there was no tcl-tk by default. I use matplotlib sometimes and the first thing it will do is to look for tcl-tk.
The mac version has tk support by default (see the Python test block on mac). We try to keep mac and linux behaviour aligned when possible.

Sometimes aligning ourselves with Homebrew on macOS make sense, and sometimes aligning ourselves with the Debian / Ubuntu approach is the better choice. In this case, I prefer the Debian / Ubuntu way. They name the packages `python` and `python-tk`.

See https://packages.debian.org/stretch/python
and https://packages.debian.org/stretch/python-tk

macOS ships with a graphical interface and Tcl/Tk, so it adds no dependencies to include Tcl/Tk by default with the brewed `python`. The situation is different on Linux where adding Tcl/Tk by default pulls in a bunch of additional dependencies.

sjackman:
>> It's only a difference of naming, but could we go the other way and add a python-tcl-tk formula?

> People (me included) expect to brew install python and get tcl-tk support. Also, python-tcl-tk would be keg-only, which makes it less nice to use.

It might be possible for `python-tk` to install only the files needed to add Tcl/Tk support to Python, specifically I'm thinking of the shared libraries with linkage to `tcl-tk`.

sjackman:
So this is interesting. On Debian, `python-tk` installs only a single shared library (plus some metafiles): `/usr/lib/python2.7/lib-dynload/_tkinter.so`
See https://packages.debian.org/stretch/amd64/python-tk/filelist
I'm thinking that we can use this same solution for Linuxbrew!

Python 3 is the same: https://packages.debian.org/stretch/amd64/python3-tk/filelist

iMichka:
We could maybe do that. There is probably a way to compile that file alone. Or compile python, install the file, and drop the rest (though it will need to be linked against the correct paths).

But in general this is not the strategy used by homebrew. We provide everything in a single formula, and try to avoid splitting stuff. This would be a new behaviour (and exception), which may also surprise some people.

iMichka:
Just to be clear: I do not want to close the door on that idea, but I am not really in favour of it.

sjackman:
>  There is probably a way to compile that file alone. Or compile python, install the file, and drop the rest

I was planning on just compiling the whole thing and dropping the rest, but good point, there's probably a way to compile just that one file.

> (though it will need to be linked against the correct paths).

Right. I guess we'd need to modify the search path of `python` to find that `_tkinter.so`. I think that can be done by modifying the site-local search path file installed by `python`. I forget the name of that file, but I believe it exists.

> But in general this is not the strategy used by homebrew. We provide everything in a single formula, and try to avoid splitting stuff. This would be a new behaviour (and exception), which may also surprise some people.

True. In this case I think it's worth the exception to avoid pulling in `xorg`, which is a large dependency, and unnecessary on a headless system, and because `python` is such a common build (and run time) dependency.

iMichka:
I am not working on this anymore.

PR #461, opened 2019-02-21
libxt: make check requires glib

sjackman:
Fix the error:
```
error: glib.h: No such file or directory compilation terminated.
```

See https://github.com/Linuxbrew/homebrew-xorg/issues/460

This PR may be squash merged without updating the bottle.

### DIFF
--- a/Formula/libxt.rb
+++ b/Formula/libxt.rb
@@ -13,9 +13,8 @@ class Libxt < Formula
   option "without-test", "Skip compile-time tests"
   option "with-static", "Build static libraries (not recommended)"
   option "with-specs", "Build specifications"
-  option "with-glib", "Build with glib (for unit testing)"
 
-  depends_on "glib" => [:build, :optional]
+  depends_on "glib" => :build if build.with "test"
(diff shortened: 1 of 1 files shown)


Formula/libxt.rb
```diff
@@ -13,9 +13,8 @@ class Libxt < Formula
   option "without-test", "Skip compile-time tests"
   option "with-static", "Build static libraries (not recommended)"
   option "with-specs", "Build specifications"
-  option "with-glib", "Build with glib (for unit testing)"
 
-  depends_on "glib" => [:build, :optional]
+  depends_on "glib" => [:build, :recommended]
```
maxim-belkin: How about
```ruby
depends_on "glib" => :build if build.with? "test"
...
system "make", "check" if build.with? "test"
```
?

sjackman: Done.

maxim-belkin (approved):

maxim-belkin:
Thank you, Shaun!

sjackman:
No worries!

PR #483, opened 2019-03-26
libxt: Make running tests optional

iMichka:
This simplifies the dependency tree: it removes glib, meson, ninja
and Python.

### DIFF
--- a/Formula/libxt.rb
+++ b/Formula/libxt.rb
@@ -10,7 +10,7 @@ class Libxt < Formula
     sha256 "7898201b493d05244375a0d092f3f642e4435ea7fd08ce95a1d8a5e288d83e44" => :x86_64_linux
   end
 
-  option "without-test", "Skip compile-time tests"
+  option "with-test", "Run compile-time tests"
   option "with-static", "Build static libraries (not recommended)"
   option "with-specs", "Build specifications"
 


iMichka:
Follow up from #461

maxim-belkin:
@iMichka, you merged without a review for me or Shaun...?

iMichka:
Yes, I wanted to make some quick progress here, I have a small times lot right now to work on this.

As this is just disabling the test, it is quite harmful. We can revert back in case you want to give this a second though. Sorry to have jumped the gun here.

maxim-belkin:
I would definitely want to review this one and the example where `glib` brings in the dependencies you mentioned because here it is a build-time dependency only.

iMichka:
If you build from source, you'll have these dependencies in the tree. As this is 'just' for the tests, I think the best way is to disable the test.

In core we do not run build-time tests (there are a few exceptions though).

There is one alternative: I introduced a package called `minimal-python` in core, which is python3 and keg only. That may be used to build meson, ninja and glib. But I am afraid that this may be a change that is too big to handle.

sjackman (commented):
I'm :+1: on reducing build time dependencies.

maxim-belkin:
>  I introduced a package called minimal-python in core,

I saw [that](https://github.com/Homebrew/linuxbrew-core/pull/12120) and I think that this is a good example of a no-review PR. The idea behind "minimal python" was (and still is) to have a python formula that would be a "dead end" in the dependency graph -- that is a node that does not depend on anything else. That's why we were installing python with `apt-get` in the first place. Currently, `minimal-python` formula is the same `python` formula with all of its dependencies. I don't see how it can help us. Even if it somehow does help - this discussion should've taken place before `minimal-python` was added.

iMichka:
I do hundreds of non-review PRs in core during the week. I move dependencies, remove stuff, make choices. Everything can be reverted or changed by pull-request if needed. It would be great if everything could always be reviewed ... I will give more details about my idea of `minimal-python` in #455, where it should be the only place it will be used right now.

The change about `minimal-python` is independent from this PR. Let's focus on the change in `libxt` here.

PR #375, opened 2018-06-08
mesa: failed with llvm@6

Gwang-Jin:
I changed `llvm@6` into `llvm@4`.
`--enable-gallium-llvm` option was deprecated and needed to be changed to `--enable-llvm.`

# Contributing to homebrew-xorg:

_You can erase any parts of this template not applicable to your Pull Request._

- [x] Have you followed the guidelines in our [Contributing](https://github.com/Linuxbrew/homebrew-xorg/blob/master/.github/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Linuxbrew/homebrew-xorg/pulls) for the same update/change?
- [x] Have you included a log of the failed build without your patch using `brew gist-logs <formula>`?
- [x] Does your submission pass
`brew audit --strict <formula>` (where `<formula>` is the name of the formula you're submitting)?
- [x] Have you built your formula locally prior to submission with `brew install <formula>`?

PR #484, opened 2019-06-12
mesa: bump version to match linuxbrew-core

CL-Jeremy:

- [ ] Have you included a log of the failed build without your patch using `brew gist-logs <formula>`?

---
I'm not sure when it happened, but the old tarball is no longer available for download.
I'm running Linuxbrew on a public login server (with home folders over NFS) so I had to build from source. Only then did I encounter this issue.
Moreover I find it hard to believe how old this package is, with even a resource (`libva`) being much more up to date as a standalone formula, in the very same tap. Just out of curiosity, has there been specific issues in the past preventing you from updating `mesa`?

### DIFF
--- a/Formula/mesa.rb
+++ b/Formula/mesa.rb
@@ -25,7 +25,7 @@ class Mesa < Formula
   depends_on "linuxbrew/xorg/xorgproto" => :build
   depends_on "llvm@7" => :build
   depends_on "pkg-config" => :build
-  depends_on "python" => :build
+  depends_on "python@2" => :build
   depends_on "libelf"
   depends_on "linuxbrew/xorg/libdrm"
   depends_on "linuxbrew/xorg/libomxil-bellagio"
--- a/Formula/mesa.rb
+++ b/Formula/mesa.rb
@@ -1,9 +1,9 @@
 class Mesa < Formula
   desc "Cross-driver middleware"
   homepage "https://dri.freedesktop.org"
-  url "https://mesa.freedesktop.org/archive/mesa-17.2.3.tar.xz"
-  sha256 "a0b0ec8f7b24dd044d7ab30a8c7e6d3767521e245f88d4ed5dd93315dc56f837"
-  revision 3
+  url "https://mesa.freedesktop.org/archive/mesa-19.0.6.tar.xz"
+  sha256 "2db2f2fcaa4048b16e066fad76b8a93944f7d06d329972b0f5fd5ce692ce3d24"
+  head "https://gitlab.freedesktop.org/mesa/mesa.git"
 
   bottle do
     root_url "https://linuxbrew.bintray.com/bottles-xorg"
@@ -23,13 +23,14 @@ class Mesa < Formula
   depends_on "linuxbrew/xorg/libvdpau" => :build
   depends_on "linuxbrew/xorg/wayland-protocols" => [:recommended, :build]
   depends_on "linuxbrew/xorg/xorgproto" => :build
-  depends_on "llvm@4" => :build
+  depends_on "llvm@7" => :build
   depends_on "pkg-config" => :build
   depends_on "python" => :build
   depends_on "libelf"
   depends_on "linuxbrew/xorg/libdrm"
   depends_on "linuxbrew/xorg/libomxil-bellagio"
   depends_on "linuxbrew/xorg/libxdamage"
+  depends_on "linuxbrew/xorg/libxrandr"
   depends_on "linuxbrew/xorg/libxshmfence"
   depends_on "linuxbrew/xorg/libxv"
   depends_on "linuxbrew/xorg/libxvmc"
@@ -47,13 +48,13 @@ class Mesa < Formula
   #
 
   resource "mako" do
-    url "https://files.pythonhosted.org/packages/56/4b/cb75836863a6382199aefb3d3809937e21fa4cb0db15a4f4ba0ecc2e7e8e/Mako-1.0.6.tar.gz"
-    sha256 "48559ebd872a8e77f92005884b3d88ffae552812cdf17db6768e5c3be5ebbe0d"
+    url "https://files.pythonhosted.org/packages/f9/93/63f78c552e4397549499169198698de23b559b52e57f27d967690811d16d/Mako-1.0.10.tar.gz"
+    sha256 "7165919e78e1feb68b4dbe829871ea9941398178fa58e6beedb9ba14acf63965"
   end
 
   resource "libva" do
-    url "https://www.freedesktop.org/software/vaapi/releases/libva/libva-1.7.3.tar.bz2"
-    sha256 "22bc139498065a7950d966dbdb000cad04905cbd3dc8f3541f80d36c4670b9d9"
+    url "https://github.com/intel/libva/releases/download/2.4.0/libva-2.4.0.tar.bz2"
+    sha256 "99263056c21593a26f2ece812aee6fe60142b49e6cd46cb33c8dddf18fc19391"
   end
 
   patch :p1 do
@@ -77,6 +78,7 @@ class Mesa < Formula
     args = %W[
       CFLAGS=#{ENV.cflags}
       CXXFLAGS=#{ENV.cflags}
+      --enable-autotools
       --disable-silent-rules
       --disable-dependency-tracking
       --prefix=#{prefix}
@@ -86,7 +88,7 @@ class Mesa < Formula
       --enable-llvm
       --disable-llvm-shared-libs
       --enable-shared-glapi
-      --with-llvm-prefix=#{Formula["llvm@4"].opt_prefix}
+      --with-llvm-prefix=#{Formula["llvm@7"].opt_prefix}
       --enable-dri3=#{gpu}
       --enable-dri=#{gpu}
       --enable-egl=#{gpu}
--- a/Formula/mesa.rb
+++ b/Formula/mesa.rb
@@ -21,6 +21,7 @@ class Mesa < Formula
   depends_on "libtool" => :build
   depends_on "linuxbrew/xorg/libpthread-stubs" => :build
   depends_on "linuxbrew/xorg/libvdpau" => :build
+  depends_on "linuxbrew/xorg/libxrandr" => :build
   depends_on "linuxbrew/xorg/wayland-protocols" => [:recommended, :build]
   depends_on "linuxbrew/xorg/xorgproto" => :build
   depends_on "llvm@7" => :build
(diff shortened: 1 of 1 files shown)


CL-Jeremy:
Just to say that this did build (cf. checklist item no. 5) on my platform without changing `python` to `python@2` due to the fact that Debian still hasn't switched to defaulting `python` -> `python3` so the system's Python 2 was used by mistake.

CL-Jeremy:
Hmm... this depends on `systemd` which opportunistically links to glib and thus also GTK & Co. (I assume this kind of thing not happen under macOS, as with `pulseaudio`, for example).
I currently have systemd built with an extra flag set:
```
diff --git a/Formula/systemd.rb b/Formula/systemd.rb
index 0ea0d350a9..641a9bf485 100644
--- a/Formula/systemd.rb
+++ b/Formula/systemd.rb
@@ -52,6 +52,7 @@ class Systemd < Formula
       "--disable-acl",
       "--disable-debug",
       "--disable-dependency-tracking",
+      "--disable-glib",
       "--disable-silent-rules",
       "--prefix=#{prefix}",
       "--localstatedir=#{var}",
```
while for `pulseaudio` I actually added `glib` dependency specifically for Linux, since the parts involving GTK are mostly needed by other formulae (though I remember there was also an option regarding `gsettings` integration, I left that out):
```
diff --git a/Formula/pulseaudio.rb b/Formula/pulseaudio.rb
index 255ba0708f..251748f2dc 100644
--- a/Formula/pulseaudio.rb
+++ b/Formula/pulseaudio.rb
@@ -33,6 +33,7 @@ class Pulseaudio < Formula
     depends_on "m4" => :build
     depends_on "libcap"
     depends_on "expat"
+    depends_on "glib"

     # Depends on XML::Parser
     # Using the host's Perl interpreter to install XML::Parser fails when using brew's glibc.
```
So I think this has to wait until [Homebrew/linuxbrew-core](https://github.com/Homebrew/linuxbrew-core) sorts out all those optional GTK stuff which have always been irrelevant under macOS.

CL-Jeremy:
@iMichka would you mind having a look over here?

maxim-belkin:
Hi, Mike. Thank you for the pull request. I'm the maintainer of this repo. 

>  Just out of curiosity, has there been specific issues in the past preventing you from updating `mesa`?

 There are several reasons this formula is so out of date but the main one is that I was waiting on something in the core tap to be fixed. I don't remember what it was as it has been a while.

I will comment on the changes in a bit.

Formula/mesa.rb
```diff
@@ -23,13 +23,14 @@ class Mesa < Formula
   depends_on "linuxbrew/xorg/libvdpau" => :build
   depends_on "linuxbrew/xorg/wayland-protocols" => [:recommended, :build]
   depends_on "linuxbrew/xorg/xorgproto" => :build
-  depends_on "llvm@4" => :build
+  depends_on "llvm@7" => :build
```
maxim-belkin, line 27: It reminds me that I was waiting on `llvm@4` formula to get a Linuxbrew bottle. Initially, I too was [using](https://github.com/Linuxbrew/homebrew-xorg/blame/4cf6cf38bd369d299b44602e932f9a8623286ae3/Formula/mesa.rb) `llvm` but then [switched](https://github.com/Linuxbrew/homebrew-xorg/commit/4a5530d25a150028648d76e730988c8574ed6790) to `llvm@4` because `llvm` was [causing](https://github.com/Linuxbrew/homebrew-xorg/pull/375) some trouble. As you can see (if you have a look at the last link), we had a plan to stop using `llvm` even then.

Formula/mesa.rb
```diff
@@ -23,13 +23,14 @@ class Mesa < Formula
...
   depends_on "pkg-config" => :build
-  depends_on "python" => :build
+  depends_on "python@2" => :build
```
maxim-belkin, line 29: 👍

Formula/mesa.rb
```diff
@@ -47,13 +48,13 @@ class Mesa < Formula
...
-    sha256 "48559ebd872a8e77f92005884b3d88ffae552812cdf17db6768e5c3be5ebbe0d"
+    url "https://files.pythonhosted.org/packages/f9/93/63f78c552e4397549499169198698de23b559b52e57f27d967690811d16d/Mako-1.0.10.tar.gz"
+    sha256 "7165919e78e1feb68b4dbe829871ea9941398178fa58e6beedb9ba14acf63965"
   end
 
   resource "libva" do
-    url "https://www.freedesktop.org/software/vaapi/releases/libva/libva-1.7.3.tar.bz2"
-    sha256 "22bc139498065a7950d966dbdb000cad04905cbd3dc8f3541f80d36c4670b9d9"
+    url "https://github.com/intel/libva/releases/download/2.4.0/libva-2.4.0.tar.bz2"
+    sha256 "99263056c21593a26f2ece812aee6fe60142b49e6cd46cb33c8dddf18fc19391"
```
maxim-belkin, line 57: 👍

Formula/mesa.rb
```diff
@@ -23,13 +23,14 @@ class Mesa < Formula
...
-  depends_on "llvm@4" => :build
+  depends_on "llvm@7" => :build
   depends_on "pkg-config" => :build
-  depends_on "python" => :build
+  depends_on "python@2" => :build
   depends_on "libelf"
   depends_on "linuxbrew/xorg/libdrm"
   depends_on "linuxbrew/xorg/libomxil-bellagio"
   depends_on "linuxbrew/xorg/libxdamage"
+  depends_on "linuxbrew/xorg/libxrandr"
```
maxim-belkin: why do we need it? Is it linked against in mesa 19? Could you please show `brew linkage mesa` when you build `mesa` locally?

maxim-belkin:
All changes look good to me. Thank you! 
We need to be careful with `llvm` as we might be going back and worth with it. If you have the time, have a look at the problem in #375 and see if switching to `llvm@7` can cause it again.

Formula/mesa.rb
CL-Jeremy: Thanks for tellng me about this command as it seems quite reliable compared to a manual check. The answer appears to be no. I'd have to recall the error that failed my build. I need to investigate the situation further.

Formula/mesa.rb
CL-Jeremy: Yup, I've read all of that thread before submitting this. The issues with `llvm` are apparently not the same as before in mesa 19, but rather due to some (widely known) breaking changes in LLVM 8, which is why the switch to `llvm@7` worked out well.

iMichka:
This looks good to me. There is still the question about `libxrandr`, but we are on the good way.

Formula/mesa.rb
maxim-belkin: OK, `libxrandr` _is_ necessary to build this new version of `mesa`:
```
Last 15 lines from /home/linuxbrew/.cache/Homebrew/Logs/mesa/02.autogen.sh:
checking for x11-xcb xcb xcb-dri2 >= 1.8 xcb-xfixes... yes
checking for x11-xcb xcb >= 1.9.3 xcb-dri3 xcb-xfixes xcb-present xcb-sync xshmfence >= 1.1... yes
checking for xcb-dri3 >= 1.13 xcb-present >= 1.13... yes
checking for x11-xcb xcb-randr... yes
checking for xrandr... no
configure: error: Package requirements (xrandr) were not met:

No package 'xrandr' found

Consider adjusting the PKG_CONFIG_PATH environment variable if you
installed software in a non-standard prefix.

Alternatively, you may set the environment variables XLIB_RANDR_CFLAGS
and XLIB_RANDR_LIBS to avoid the need to call pkg-config.
See the pkg-config man page for more details.
/usr/bin/curl --version
```
but `mesa` is not linked against `libxrandr`. This means that it should be a `:build` dependency unless some paths related to `libxrandr` are hard-coded into `mesa` files...

maxim-belkin: `grep -r libxrandr *` in `mesa`'s cellar as well as grepping `libxrandr` from the output of `strings` returned nothing, so I think it is safe to set `libxrandr` as a `:build` dependency.

iMichka (approved):

CL-Jeremy:
Sorry, but upon reinstallation I don't seem to find the evidence where `configure` failed complaining about the absence of `xrandr`...
It's on this line (the test was negative in the beginning):
https://gist.github.com/CL-Jeremy/641c77b7f16ccedc49e105df5d9f9918#file-02-autogen-sh-L359
This build went through without `depends_on "linuxbrew/xorg/libxrandr"`. I was pretty speechless.

CL-Jeremy:
> upon reinstallation I don't seem to find the evidence where configure failed complaining about the absence of `xrandr`...

Okay, now I'll have to remember to uninstall the formula in question as well...
Here you go: https://gist.github.com/4910ecc741ce599f6b20cef18b8d9621

CL-Jeremy:
I reran the installation in interactive mode and tried to find out what exactly needed `xrandr`.
`grep "xrandr" -r .` (later also `grep "randr" -r .` but still no direct reference to `xrandr`'s headers:
```
./src/amd/vulkan/meson.build:  radv_deps += [dep_xcb_xrandr, dep_xlib_xrandr]
./src/intel/vulkan/meson.build:  anv_deps += [dep_xcb_xrandr, dep_xlib_xrandr]
./src/vulkan/wsi/meson.build:  vulkan_wsi_deps += [dep_xcb_xrandr, dep_xlib_xrandr]
./src/vulkan/registry/vk.xml:        <platform name="xlib_xrandr" protect="VK_USE_PLATFORM_XLIB_XRANDR_EXT" comment="X Window System, Xlib client library, XRandR extension"/>
./src/vulkan/registry/vk.xml:        <extension name="VK_EXT_acquire_xlib_display" number="90" type="instance" requires="VK_EXT_direct_mode_display" author="NV" contact="James Jones @cubanismo" platform="xlib_xrandr" supported="vulkan">
```
This justifies `=> :build` to me 👍

maxim-belkin:
> Okay, now I'll have to remember to uninstall the formula in question as well...

👍 yep

Formula/mesa.rb
CL-Jeremy: Oh my, I did not notice your replies on this comment 😯 but good that we've come up with similar conclusions.

maxim-belkin: no worries!

CL-Jeremy:
Tried with `--with-vulkan-drivers=intel,radeon` and even so `libxrandr` is not linked.
Bildschirmfoto 2019-06-18 um 10 50 46
An interesting comparision with Debian Buster (downloaded, not installed) showing LLVM 7 linkage:
Bildschirmfoto 2019-06-18 um 10 06 44
Stretch links to LLVM 3.9:
Bildschirmfoto 2019-06-18 um 10 49 51
There seems to be quite some room of improvement regarding linkage and stuff, but it's already really nice to have stuff working without root, from bottles as well as from source.

Edit: just checked: our `vdpau` drivers also do not link to libLLVM, while Debian ones do...

maxim-belkin:
> just checked: our `vdpau` drivers also do not link to libLLVM, while Debian ones do...

We don't use LLVM when building `libvdpau` so it would've been strange if it linked against it.

maxim-belkin:
Thank you, @CL-Jeremy, for contributing to Linuxbrew/xorg! 🎉 
I hope you'll keep poking this and other formulae to keep them fresh!

PR #537, opened 2019-07-22
intel-gmmlib: new formula

maxim-belkin:
New formula


maxim-belkin:
Failure due to missing test block. Ignoring for now.

PR #630, opened 2020-03-26
xkeyboardconfig: migrate to core

maxim-belkin:


maxim-belkin:
Shaun suggested migrating all formulae at once, so closing this.

PR #645, opened 2020-05-23
wayland: rebuild bottle

maxim-belkin:


iMichka:
Superseded by #646

maxim-belkin:
Homebrew/brew currently builds bottles in which binaries and libraries rely on glibc 2.14, 2.15, and 2.17. Homebrew/brew claims to support systems with glibc 2.13 so bottles that Homebrew builds shouldn't rely on stuff newer than 2.13.

iMichka:
> Homebrew/brew currently builds bottles in which binaries and libraries rely on glibc 2.14, 2.15, and 2.17. Homebrew/brew claims to support systems with glibc 2.13 so bottles that Homebrew builds shouldn't rely on stuff newer than 2.13.

This issue predates the docker container changes we made lately. There are many bottles affected by this. I merged this PR, this bottle is compatible with ubuntu 16.04 at least. We will need to see how we can solve the issue in a more general way for all the bottles.

maxim-belkin:
There were some old libffi symbols in wayland so wayland bottle had to be rebuilt. Then bad glibc symbols crippled into the new bottle because of incorrect container used for building that bottle. Then, when everything was fixed, the wayland bottle was updated (this PR).

PR #652, opened 2020-05-31
.mailmap: Bob's name

maxim-belkin:
@rwhogg: what do you think about adding a `.mailmap` file so that `git shortlog -ns` combines your contributions under one name?

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1,1 +1,2 @@
-Bob W. Hogg <wombat@rwhogg.site>
+Bob "Wombat" Hogg <rwhogg@linux.com>
+Bob "Wombat" Hogg <wombat@rwhogg.site>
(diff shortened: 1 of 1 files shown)


bw-yuhogg:
I'm :+1: on combining them (I have gone through a fair share of commit names and email addresses...). However, may I request something like this instead?

```
diff --git a/.mailmap b/.mailmap
index 316de54..89e53b9 100644
--- a/.mailmap
+++ b/.mailmap
@@ -1 +1,3 @@
-Bob W. Hogg <wombat@rwhogg.site>
+Bob "Wombat" Hogg <rwhogg@linux.com>
+Bob "Wombat" Hogg <wombat@rwhogg.site>
+
```

The full "Wombat" form is what is in my current .gitconfig; the reason it needs both lines is some Git tools I use respect the quote marks and some seem to silently drop them.

PR #657, opened 2020-06-23
Libx11 test

maxim-belkin:


maxim-belkin:
Failures are unrelated to the test.

PR #660, opened 2020-07-04
Update Xorg to current LivecheckStrategy format

samford:

---

Homebrew/homebrew-livecheck#1025 is now merged, so I'm updating the `Xorg` strategy here to the reflect the current `LivecheckStrategy` format.

Namely, there were some changes around `NAME` and `NICE_NAME` between the time this `Xorg` strategy was created and the aforementioned PR was merged:

* `NAME` should now always be `NAME = name.demodulize` (like the built-in strategies)
* `NICE_NAME` should only be provided when it differs from the class name. Using `Sourceforge` as an example, the class name has to be `Sourceforge` (so the strategy symbol is `:sourceforge`) and we set `NICE_NAME = "SourceForge"` to make the nicer name available. We currently don't use `NICE_NAME` anywhere inside livecheck, so it may be removed in the future but I'm making it available for now.

### DIFF
--- a/livecheck_strategy/xorg.rb
+++ b/livecheck_strategy/xorg.rb
@@ -2,8 +2,7 @@
 
 module LivecheckStrategy
   class Xorg
-    NICE_NAME = "Xorg"
-    NAME = NICE_NAME.downcase
+    NAME = name.demodulize
 
     def self.match?(url)
       /(...\.x\.org|freedesktop\.org)/.match?(url)


maxim-belkin:
Thanks for keeping this tap up-to-date with Homebrew-livecheck, Sam!

>     * `NAME` should now always be `NAME = name.demodulize` (like the built-in strategies)

Given that it has to be the same for all strategies, I feel that it should live somewhere in Homebrew/livecheck, not in strategy definition (unless it is used to override automatically-computed value).

samford:
> I feel that it should live somewhere in Homebrew/livecheck, not in strategy definition

Since it's always the same code in every strategy, I agree that creating a solution in livecheck itself would be ideal from the standpoint of strategy authors.

My reasoning here was that having `name.demodulize` in the strategy itself ensures it's only executed once per strategy across a livecheck run. This was to avoid the situation where we would have `strategy_name = strategy.name.demodulize` in `latest_version()`, where it would be executed for each URL that livecheck tries for each formula. Over the course of a full livecheck run, this would be unnecessarily executed several thousand times. Simply requiring that each strategy have the same `NAME` constant was the easiest way to deal with this at the time.

That said, I implemented a simple solution just now that removes the `NAME` requirement (through some additional code in the livecheck command), while also achieving the same result (only executing `name.demodulize` once per strategy per run). I'll create a PR for the changes once I get through some other important ones in the coming days. Once that's done and merged, I'll come back and remove `NAME` from the `Xorg` strategy here.

PR #661, opened 2020-07-04
Integrate livecheckables into formulae

samford:

---

The `brew livecheck` command will be migrated from Homebrew/homebrew-livecheck into Homebrew/brew in the not-too-distant future and we will be removing the `Formulary` extension that enables the use of files in a `Livecheckables` folder in the process.

The reason for this is that livecheckables will be integrated into formulae as a `livecheck` block. The current `livecheck` block format in livecheckables is a half-step in between. Livecheck has supported `livecheck` blocks in formulae for some time (and favors it over a livecheckable file) but we haven't yet migrated livecheckables in homebrew-livecheck to homebrew-core formulae as we need to take care of some other things first. [We're currently bringing existing livecheckables (namely, the older ones) up to existing standards before the migration.]

With all this in mind, there's no reason not to integrate the livecheckables in this repository into their corresponding formulae. Doing this now will prevent them from breaking after the Homebrew/brew migration, so I thought I would take a moment to take care of this.

Besides that, I also modified the `regex` calls to always use parentheses, since that's what we've standardized on for technical reasons.

I've confirmed that all checks for this tap work exactly the same both before and after these changes. Anything that worked before continues to work and anything that was broken is still broken (in the same manner).

### DIFF
--- a/Formula/intel-media-driver.rb
+++ b/Formula/intel-media-driver.rb
@@ -4,6 +4,11 @@ class IntelMediaDriver < Formula
   url "https://github.com/intel/media-driver/archive/intel-media-19.4.0r.tar.gz"
   sha256 "a03bd75eefe9cb0245e3aab2723b3fef555d9f180a180b2c29d7b12d483d9ec2"
 
+  livecheck do
+    url "https://github.com/intel/media-driver/releases"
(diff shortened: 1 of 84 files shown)


maxim-belkin:
Thank you, Sam! Excellent contribution -- much appreciated!

PR #662, opened 2020-07-11
xdpyinfo: Add a test

rwhogg:
Also delete some commented-out code

### DIFF
--- a/Formula/xdpyinfo.rb
+++ b/Formula/xdpyinfo.rb
@@ -26,13 +26,12 @@ class Xdpyinfo < Formula
       --disable-dependency-tracking
       --disable-silent-rules
     ]
-    #  --without-dga           Disable dga support.
-    #  --without-xf86misc      Disable xf86misc support.
-    #  --without-xinerama      Disable xinerama support.
-    #  --without-dmx           Disable dmx support.
-
     system "./configure", *args
     system "make"
     system "make", "install"
+  end
+
+  test do
+    assert_match /version number:\s+11.0/, shell_output("xdpyinfo")
   end
 end


bw-yuhogg:
Travis failed:

```
$ brew audit -D --except=revision_and_version_scheme --strict $(cd $(brew --repo linuxbrew/xorg)/Formula; ls *.rb | sed 's|.rb$||') || true
linuxbrew/xorg/libevdev:
  * Dependency 'python' from homebrew/core is an alias; use the canonical name 'python@3.8'.
linuxbrew/xorg/libxcb:
  * Dependency 'python' from homebrew/core is an alias; use the canonical name 'python@3.8'.
linuxbrew/xorg/xcb-proto:
  * Dependency 'python' from homebrew/core is an alias; use the canonical name 'python@3.8'.
```

I'll admit that I have not kept pace with all the changes to Python that have happened recently - if someone can confirm that we do want `python@3.8` rather than the nonspecific `python` I can make the above changes on a separate PR but at the moment I'm not sure what the proper course of action is.

jonchang:
Yeah, we've moved towards versioning each minor release of Python. Users running `brew install python` will automatically install `python@3.8`, and should automatically upgrade to `python@3.9` when that releases, but formulae depending on `python@3.8` will have more time to migrate to e.g. `python@3.9` without holding up user access to the newer version of Python.

> if someone can confirm that we do want python@3.8 rather than the nonspecific python I can make the above changes on a separate PR but at the moment I'm not sure what the proper course of action is.

This is exactly what we want! If you can spare the time to open the pull request that would be fantastic.

bw-yuhogg:
Thanks for explaining, that makes a lot of sense :+1:

Opened up a :snake: PR at #663

maxim-belkin (approved):
Tested locally -- works great! Thanks, Bob!

PR #663, opened 2020-07-12
xcb-proto, libxcb, libevdev: depend on Python@3.8

rwhogg:

`brew audit --strict <formula>` (where `<formula>` is the name of the formula you're submitting)? **No. I did add a test for `xcb-proto` but not for `libxcb`, which will still fail until it gets one.**

Following up on the audit failure I found on #662 . A few things worth noting:

* I bumped the revision on `xcb-proto` because its existing bottle had a python3.7 path.
* I also used the `PYTHON` environment variable to enforce python choice, as on my machine the initial result on building from source was python2.7
* Since `xcb-proto` got a revision bump, I decided `libxcb` should get one too
* `libevdev` on the other hand didn't get a revision bump, as it doesn't depend on either of the other two and I didn't see any Python references in the resulting installed directory (other than a URL).
* I'm assuming that anything that depends on `libxcb` should be rebuilt automatically on upgrade if necessary. So I didn't bump the revisions of stuff like `xcb-util-renderutil`. (I'm admittedly a little fuzzy on whether those will need rebuilds eventually or not).

Feel free to overrule me on any of these decisions.

### DIFF
--- a/Formula/libevdev.rb
+++ b/Formula/libevdev.rb
@@ -15,7 +15,7 @@ class Libevdev < Formula
 
   depends_on "check" => :build if build.with? "test"
   depends_on "pkg-config" => :build
-  depends_on "python" => :build
+  depends_on "python@3.8" => :build
 
   def install
     args = %W[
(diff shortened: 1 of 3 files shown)


maxim-belkin:
Thanks for helping out with the tap, Bob! I wrote a test for libxcb in #664.

PR #665, opened 2020-07-13
Fix audits

jonchang:
https://github.com/Homebrew/brew/pull/7983 unmasked a years-old bug in Homebrew where `depends_on ... if build.with?` wasn't properly getting checked by `brew audit`. This pull request fixes those audits in Linuxbrew/xorg.


maxim-belkin (approved):
Thanks, @jonchang!

PR #678, opened 2020-07-15
xorg, xorg-apps, xorg-fonts: add livecheck block

samford:

This adds `livecheck` blocks to skip `xorg`, `xorg-apps`, and `xorg-fonts`, since they only install other formulae and they can't be checked like normal formulae. Let me know if you would prefer a different skip message and I'll update these.

### DIFF
--- a/Formula/xorg-apps.rb
+++ b/Formula/xorg-apps.rb
@@ -7,6 +7,10 @@ class XorgApps < Formula
   sha256   "d6f8969d76bd3236c40e57fc3f498c43341f3f1e9ae01c4c2bca11876f07adc6"
   # tag "linuxbrew"
 
+  livecheck do
+    skip "Formula only installs other formulae"
+  end
+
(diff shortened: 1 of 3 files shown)


maxim-belkin (approved):
Thank you, Sam 7!

samford:
Between the individual `livecheck` block PRs and the improvements to the `Xorg` strategy, livecheck now works for all the formulae in this tap 🎉

`brew livecheck --tap linuxbrew/xorg` is also dramatically faster now, since I added some simple page caching in the `Xorg` strategy (primarily to avoid checking the same index pages more than once).

maxim-belkin:
> `brew livecheck --tap linuxbrew/xorg` is also dramatically faster now, since I added some simple page caching in the `Xorg` strateg

Thanks, Sam. I can confirm that `brew livecheck --tap linuxbrew/xorg` is definitely faster! 🎉

PR #679, opened 2020-07-29
Fix style

maxim-belkin:


jonchang:
Think the workflow needs a `brew update` to get the latest version of Homebrew.

maxim-belkin:
Yes, looks like it...

PR #844, opened 2026-03-29
xset 1.2.5

maxim-belkin:


maxim-belkin:
Closing: need to rebuild dep bottles (libxfontcache, libxxf86misc) first.

PR #846, opened 2026-03-29
libxxf86misc: rebuild bottle

maxim-belkin:
Rebuild bottle for compatibility with current CI environment.


maxim-belkin:
CI can't produce bottles due to dep compatibility. These formulas need manual bottle builds.

PR #863, opened 2026-04-01
libxxf86misc: add bottle

maxim-belkin:


maxim-belkin:
libxxf86misc requires xf86miscproto which is unavailable. Dropping libxxf86misc from xset instead.

